PR #130, opened 2025-07-01
Update README.md with step to add general registry on fresh Julia install

jeanettejohnson:
Trying to add the PCVCT registry on a fresh Julia install without installing the general registry first gives a registry not-found error.

See here: https://forum.mimiframework.org/t/error-installing-mimi-under-v1-3-1/109/4

Error: (@v1.11) pkg> add pcvct
    Updating registry at `~/.julia/registries/PCVCTRegistry`
    Updating git-repo `https://github.com/drbergman/PCVCTRegistry`
   Resolving package versions...
ERROR: cannot find name corresponding to UUID 995b91a9-d308-5afd-9ec6-746e21dbc043 in a registry

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@ Make sure you are familiar with the [Best practices](https://drbergman.github.io
 See [Getting started](https://drbergman.github.io/pcvct/stable/man/getting_started/) for more details.
 
 1. [Install Julia](https://julialang.org/install).
-2. Add the general registry:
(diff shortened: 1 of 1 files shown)


jeanettejohnson:
Oh definitely! And yeah it's weird, I think it's a pretty unique error to
literally the first package you try to install so it likely won't be a
problem for most anyway

Jeanette Johnson

PR #221, opened 2026-07-31
Prevent re-initialization of the PhysiCellModelManager.jl project in …

drbergman:
…__init__ function

### DIFF
--- a/src/PhysiCellModelManager.jl
+++ b/src/PhysiCellModelManager.jl
@@ -62,7 +62,7 @@ struct PCMMMissingProject <: Exception
 end
 
 function __init__()
-    isInitialized() && return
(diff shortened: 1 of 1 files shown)


drbergman:
Superseded by #224.

The guard added here cannot fire in the case it was written for: `mm_globals_ref` is a `const Ref` initialized to `nothing` and PCMM's own `__init__` is its only assignment site, so it is always `nothing` in the fresh precompilation subprocess where the duplicate banner is actually produced. Measured directly — every reproduced duplicate had `ref == nothing` on entry.

#224 gates on `jl_generating_output` instead, and has been retargeted at `main` so the add-then-remove churn stays out of the history.

PR #227, opened 2026-08-19
Make the compile decision honest: name the executable for the PhysiCell version, and re-check that version every time

drbergman:
Three commits, one theme: every record PCMM used to decide "is there a usable build?" could disagree with reality.

## 1. A compile error looked like a finished build

`loadCustomCode` decided from two records:

- `physicell_commit_hash.txt`, holding the PhysiCell version — **written before `make` ran**
- the executable, always named `project`

A compile error left the old binary in place beside a file now claiming the *new* version. The next run read a matching hash, found `project`, concluded nothing had changed, and silently ran a binary built from different PhysiCell source.

**Fix:** the executable is named `pcmm_build/project_<physicell-version>`, so its existence **is** the record that a build for that version finished. `physicell_commit_hash.txt` is gone. A compilation that does not finish deletes any executable for that version, including one that predated it, so the next run recompiles rather than trusting what is left.

Dirty/download still recompile every run — those suffixes mean the source can change without the recorded version changing, so the name proves nothing. That is now its own predicate (`unreproduciblePhysiCellVersion`) instead of a fall-through of the hash comparison. Also: `make` exiting 0 without producing an executable is a failure return rather than an unhandled `mv` error.

Executables live in a `pcmm_build/` subfolder rather than beside `main.cpp`, so clearing them is an exact operation on a directory PCMM owns instead of a name match in a folder the user also keeps their own files in. Every other artifact name is matched exactly, so `projectile.cpp` and `project_notes.md` are both safe.

## 2. `macros.txt` had the same defect

`addMacrosIfNeeded` appended to it *before* compiling. A macro change sets both `recompile` and `clean` — macros are part of every translation unit, so object files compiled without them must be discarded. If that compilation failed, `macros.txt` already listed the new macro, so the retry saw no change and skipped the `make clean` it needed.

`neededMacros` now computes the list in memory, `compilerFlags` compares it against the file, and `writeMacrosFile` runs only after `make` succeeds. Two bugs fell out of the restructure rather than being hunted:

- **The macro checks were chained with `||`**, so a newly needed PhysiECM macro short-circuited the RoadRunner check entirely. A project needing both got one macro, failed to compile, and only picked up the second on the next attempt. The comment above the line — *"julia's `|=` operator seems to evaluate the RHS even if the LHS is already true, but I don't trust that will always be the case"* — shows the `||` rewrite defeated its own intent.
- **`addPhysiECMIfNeeded` called `addMacro(M, ...)` with an undefined `M`**, an `UndefVarError` waiting on any project that enables `ecm_setup` in config without an `ic_ecm` input. Reachable in the current test data (`isPhysiECMInConfig` returns `true` for `template-ecm`), just masked by the `ic_ecm` short-circuit ahead of it.

## 3. The PhysiCell version was only ever checked at initialization

`simulator().current_version_id` is assigned in exactly one place — during database init — and never revisited, and `physiCellCommitHash()` reads the database row for that cached id. So it cannot see the working tree. Edit `PhysiCell/Makefile` mid-session, no re-init:

```
gitDirectoryIsClean()            = false      <- PCMM can see it is dirty
physiCellCommitHash()            = b4264ae6…  <- but no -dirty suffix
unreproduciblePhysiCellVersion() = false
executableExists()               = true
loadCustomCode()                 = true       <- returned without compiling
```

Reusing the stale binary is the mild half. The damaging half: a recompile forced for any *other* reason builds the **new** source while naming the executable and writing `physicell_version_id` for the **old** version — the database then asserts a clean commit for a binary that is not that commit.

**Fix:** `refreshPhysiCellVersion()` is the first statement of `loadCustomCode`. Everything else falls out of §1, which is why these belong in one PR — a moved HEAD gives a name that does not exist yet, a newly dirty tree makes `unreproduciblePhysiCellVersion` true, and ModelManager records simulations from the same field with setup ahead of recording.

It runs per sampling, not per simulation — one `git status --porcelain`, measured 30 ms on the steady-state path — so a sampling cannot straddle two versions. `resolvePhysiCellVersionID` and `gitDirectoryIsClean` gained `verbose=true` kwargs so the refresh resolves quietly and reports one line only when the version moved; otherwise a permanently dirty PhysiCell would print its modified-file listing once per *sampling* instead of once per session.

## What you will notice

- **One extra recompile per custom code folder** after upgrading, since no `project_<version>` file exists yet. After it succeeds, `project` and `physicell_commit_hash.txt` are deleted. Self-healing, so no `up.jl` milestone — that would have dragged the whole `continueMilestoneUpgrade` prompt in to delete two files.
- **`pcmm_build/` can accumulate one binary per PhysiCell version used there.** Deliberate: the name is a cache key now, so switching versions back and forth no longer forces a rebuild. `clearSimulatorArtifacts` (reached by `resetDatabase`) deletes the whole folder. Pruning on each successful compile was rejected — it could delete a binary another session on a shared filesystem is about to launch.
- **A failed compilation now costs you the previous build too.** You asked for a rebuild and it did not happen, so there is no binary to trust; fix the compile and run again.
- **You no longer need to restart Julia** after updating PhysiCell. New `best_practices.md` sections cover doing it between campaigns and keeping the working tree clean.
- `PROGRAM_NAME` is now the final executable name; it used to build `project_ccid_<id>` and rename on the way out, which the `randstring(10)` temp directory already made pointless.

## What this does not fix

The name carries no OS, architecture, or compiler-flag component, so copying a `data/` folder to a different platform still reuses an unrunnable binary. Now written down in `known_limitations.md` with the `force_recompile=true` workaround.

Worth noting for the deferred `-march` work (`progress.md`, `2026-08-05`): that entry called for "keying the artifact on the march flag / detected ISA instead of overwriting a single `project`." The keying mechanism now exists — making `native` safe becomes a matter of adding a march/ISA component to `executableName`. Two now-stale sentences in that entry are annotated.

## `strict_check` deleted

`PhysiCellSimulator.strict_check` was documented on an **exported** type as *"If `true`, require a clean git directory to skip recompile"*, was set to `true`, and was read nowhere. `git log -S` shows it was never wired under that name or its predecessor `strict_physicell_check` — declared and forgotten, not a refactor casualty, while its docstring rendered on the site as a promise the code did not keep.

The behaviour it described already exists unconditionally in `unreproduciblePhysiCellVersion()`, so the field was never missing behaviour — it was a missing *off-switch*. Wiring it up would have meant offering to record a clean commit hash for a tree that is not that commit, in the package whose job is reproducible bookkeeping; PCMM already prints the better answer when it sees a dirty repository. Deleted. Technically breaking for nine-argument positional construction, which nothing does.


src/compilation.jl
```diff
@@ -70,7 +78,20 @@ function loadCustomCode(S::AbstractSampling; force_recompile::Bool=false)
...
-    mv(joinpath(temp_physicell_dir, executable_name), joinpath(path_to_input_custom_codes, baseToExecutable("project")), force=true)
+    path_to_compiled_executable = joinpath(temp_physicell_dir, executable_name)
+    if !isfile(path_to_compiled_executable)
+        println("""
+        Compilation exited without error, but produced no executable at $(path_to_compiled_executable).
+        Check $(joinpath(path_to_input_custom_codes, "compilation.log")) for more information.
+        """
+        )
+        rm(temp_physicell_dir; force=true, recursive=true)
+        return false
```
drbergman: Is it possible to get here with an executable for the current PhysiCell version still in the custom_codes folder? Should we also make sure to attempt the deletion of this file if it exists?

src/compilation.jl
```diff
@@ -232,9 +232,14 @@ Whether `filename`, a bare file name in a custom code folder, is a file PCMM wri
 
 Covers the executables (`project_<physicell-version>`, plus `.exe` on Windows), the
 compilation logs, `macros.txt`, and what PCMM v0.3.3 and earlier left behind: an executable
-named `project` and a `physicell_commit_hash.txt`.
+named `project` and a `physicell_commit_hash.txt`. The legacy names are matched on either
+platform so that a data folder built elsewhere is still cleaned.
+
+Matched on the `project_` prefix rather than `project`, so an unrelated `projectile.cpp` is
+left alone. A file the user happened to name `project_notes.md` is not distinguishable from
+an executable and would be deleted; nothing but generated files belongs here.
```
drbergman: this is a weird example to protect against `projectile.cpp`. you immediately undercut this "robustness" with your `project_notes.md` example. cut the extra text here. I really don't want to accidentally delete something a user puts here for their own records or whatever. so if we need to rename the compiled binary, I'd rather do that then rely on something that could match over-broadly

src/compilation.jl
drbergman: Yes, it is possible, and in one case the hole was live.

The failure paths are reached whenever `recompile` was set, and only one of the four reasons implies the executable is absent:

| trigger | executable for this version present? |
|---|---|
| `!executableExists` | no |
| macros changed | **yes, possibly** — built with the old macros |
| dirty PhysiCell | **yes, possibly** — built from a different dirty tree, same `-dirty` name |
| `force_recompile=true` | **yes, possibly** |

The first three self-heal, because the thing that triggered the recompile is still true on the next run. `force_recompile` does not: edit your custom code, run with `force_recompile=true`, hit a syntax error, then run again without the flag — unchanged macros, clean repository, executable present, so PCMM runs the binary from before your edit. Same bug as the one this PR is about, on a different axis.

Fixed in 164aca61e: `abandonBuild` deletes it on both failure paths.

I considered deleting it up front instead, before `make`, which would additionally cover a `kill -9` mid-compile. Rejected — it removes the binary for the whole duration of every rebuild, where the closing `mv` is an atomic swap, so a concurrent session on a shared filesystem would fail to launch simulations for minutes rather than not at all. The residual is that an abnormal termination can still leave a stale binary; that one at least fails loudly at launch rather than silently returning wrong results. Say the word if you want it up front anyway.

src/compilation.jl
drbergman: Agreed on both counts — the paragraph was arguing for robustness and conceding it in the next sentence, and you are right that a glob which deletes can always swallow something a user put there.

Renaming alone cannot fix it: any name carrying the version is a pattern match unless the set is enumerable. So in 164aca61e the executables moved into a `pcmm_build/` subfolder instead. `clearSimulatorArtifacts` deletes that directory outright and matches every remaining name **exactly**:

```julia
isCompilationArtifact(filename) = filename in ("compilation.log", "compilation.err",
    "macros.txt", "project", "project.exe", "physicell_commit_hash.txt")
```

No prefixes anywhere, so `projectile.cpp`, `project_notes.md`, and anything else you keep beside your `main.cpp` are safe. Verified against all 11 custom code folders in the test project: build folders and exact-name artifacts gone, planted `project_notes.md` / `projectile.cpp` / `projects.txt` untouched. The docstring paragraph you flagged is gone, and `createDefaultGitIgnore` now ignores `pcmm_build/` rather than `project*`.

A custom code folder ends up as:

```
0_template/
├── Makefile
├── main.cpp
├── custom_modules/
├── compilation.log
├── macros.txt
└── pcmm_build/
    └── project_b4264ae6d4e83d848d3d628247571e41d29ef746
```

I left `compilation.log` / `compilation.err` / `macros.txt` at the top level — exact names, no matching risk, and the paths PCMM prints in its messages stay as they are. Happy to move them in too if you would rather have one place for everything generated.

src/compilation.jl
```diff
@@ -54,6 +54,8 @@ function loadCustomCode(S::AbstractSampling; force_recompile::Bool=false)
     end
 
     executable_name = executableName()
+    path_to_build_folder = joinpath(path_to_input_custom_codes, build_folder_name)
+    path_to_executable = joinpath(path_to_build_folder, executable_name)
```
drbergman: why don't these use the functions defined below `buildFolder` and `pathToExecutable`?

src/deletion.jl
```diff
@@ -3,13 +3,23 @@
...
+there, along with the compilation logs (`compilation.log`, `compilation.err`) and the macros
+file (`macros.txt`), from each subdirectory of the custom code location.
+
+Nothing is matched by pattern: the build folder is a directory PCMM owns and everything else
+is named exactly (see `isCompilationArtifact`), so a file the user keeps alongside their
+`main.cpp` is never at risk.
 """
 function ModelManager.clearSimulatorArtifacts(::PhysiCellSimulator)
     for custom_code_folder in (readdir(locationPath(:custom_code), sort=false, join=true) |> filter(x -> isdir(x)))
+        path_to_build_folder = joinpath(custom_code_folder, build_folder_name)
```
drbergman: again, you wrote a `buildFolder` helper. why not use it?

src/compilation.jl
drbergman: No good reason — I had `path_to_input_custom_codes` in hand and reached for it, which put the layout in two places. Fixed in 78ef69723:

```julia
path_to_executable = pathToExecutable(S)
executable_name = basename(path_to_executable) #! `make` must produce exactly the file we move into place
...
mkpath(buildFolder(S))
mv(path_to_compiled_executable, path_to_executable, force=true)
```

Deriving the name from the path is a small bonus: `PROGRAM_NAME` and the destination file can no longer disagree, where before each came from its own `executableName()` call.

I also collapsed the `compilation.log` / `compilation.err` paths into two locals while I was in there — they were being rebuilt eight times across the `make` invocation, the failure message, and the empty-file check — and gave `readMacrosFile` / `writeMacrosFile` a shared `pathToMacrosFile`, since those two were assembling the same path as well.

src/deletion.jl
drbergman: This one had an actual obstacle, though not a good enough one. The loop was over full paths from `readdir(...; join=true)`, and `buildFolder` takes a folder *name* — it calls `locationPath(:custom_code, ...)` on its argument. Passing a full path would have quietly appeared to work, since `joinpath("/a/b", "/c/d")` returns `/c/d`, which is worse than failing.

So in 78ef69723 the loop iterates names instead and resolves them, matching the idiom already used for rulesets collections in `up.jl`:

```julia
custom_code_folders = readdir(locationPath(:custom_code); sort=false) |> filter(x -> isdir(locationPath(:custom_code, x)))
for custom_code_folder in custom_code_folders
    path_to_build_folder = buildFolder(custom_code_folder)
    ...
    path_to_custom_code_folder = locationPath(:custom_code, custom_code_folder)
```

Re-verified across all 11 custom code folders: build folders and exact-name artifacts gone, planted `project_notes.md` / `projectile.cpp` / `projects.txt` untouched. Full suite 617 pass with the same 5 known network failures, `CompilationTests.jl` 50/50, docs build clean.

PR #228, opened 2026-09-03
Migrate to ModelManager 0.9 and clear the release backlog

drbergman:
Brings PCMM current with in-dev ModelManager 0.9 and clears the backlog standing in front of the next release. Of the seven listed items, two were already implemented (executable naming and the PhysiCell-version re-check both landed in #227) and one — PhysiPKPD inputs — is deferred to a to-do, since it needs a schema design and every other item was independent of it. Two bugs not on the list surfaced along the way.

**This PR cannot go green until ModelManager 0.9 is released.** Both `Project.toml` and `docs/Project.toml` now require `ModelManager = "0.9"`, which is not yet registered.

## Breaking changes

Three, all narrow but worth knowing before merge.

**`runStudio` throws `PCMMStudioLaunchError`, not `Base.IOError`.** Anything catching `Base.IOError` around `runStudio` needs updating. This is the fix, not a side effect — see below.

**`configPath` rejects unrecognized motility and chemotaxis tokens.** Calls that previously returned a path now raise `ArgumentError`. Every such path was already broken; it just failed later, inside the XML layer, instead of at the call.

**`plotbycelltype` returns different numbers for monads with missing output.** The old numbers were wrong. Details below.

## `plotbycelltype` divided by the wrong denominator

The one change here that silently altered results rather than erroring.

`plotbycelltype` does not go through `MonadPopulationTimeSeries`; it re-implements the aggregation. It sized its count arrays by `monad_length = length(simulation_ids)` — the *unfiltered* replicate count — then filtered `sptss` for `missing` and filled one column per replicate that survived:

```julia
monad_length = length(simulation_ids)
sptss = SimulationPopulationTimeSeries.(simulation_ids; ...)
filter!(!ismissing, sptss)
...
cell_count_arrays[k] = zeros(Int, length(time), monad_length)
cell_count_arrays[k][:, i] = ...          # i indexes the *filtered* list
...
cell_count_means[name] = mean(array, dims=2) |> vec
```

A pruned replicate therefore left an all-zero column that the mean still divided by. Measured on a monad with one of three replicates pruned:

```
plotted (before)      = [46.7, 44.7, 44.0, 44.0]
mean over 2 survivors = [70.0, 67.0, 66.0, 66.0]
```

A one-third understatement, with no error and no warning. The denominator now comes from `length(sptss)` after filtering, which also makes it agree with `MonadPopulationTimeSeries`. `plot(::AbstractMonad)` was checked and is unaffected — it delegates to `MonadPopulationTimeSeries`, which drops missing replicates correctly.

This is why the new `@info` calls are inlined at each of the four aggregation sites rather than factored into a helper: the sites do not share an implementation, so they did not share a bug either.

## Studio error handling

`quietRun` is `run(pipeline(cmd, ...))`, which raises two unrelated types:

| failure | type | has `.code`? |
|---|---|---|
| interpreter cannot be spawned (bad `PCMM_PYTHON_PATH`) | `Base.IOError` | yes |
| Studio runs and exits non-zero | `ProcessFailedException` | **no** |

`executeStudio` built its return value as `Base.IOError(msg, e.code)`, so the second case raised `FieldError: type ProcessFailedException has no field code` from inside the error handler and the user never saw why Studio failed. It now returns a `PCMMStudioLaunchError` carrying the command and the underlying cause, so a caller can still distinguish the two.

The path had never been tested: every existing case passed a nonexistent interpreter and so took the `IOError` branch. It is covered now.

## Typed exceptions

`PCMMException` is now the supertype of every PCMM-specific error, so Model Manager Studio can catch the family or a concrete type. `PCMMMissingProject` moves under it into the new `src/exceptions.jl`; adding a supertype does not break existing `catch e isa PCMMMissingProject`.

Only two concrete types exist. A third was written for "selected rulesets folder missing its CSV" and then removed as unreachable: `basename` for that location is the vector `["base_rulesets.csv", "base_rulesets.xml"]`, so `InputFolder`'s constructor already refuses a folder containing neither, and by the time `prepareBaseFile` runs at least one exists.

*For the ModelManager side:* that `InputFolder` rejection **is** reachable by ordinary user error — create a rulesets folder, forget the rules file — and it raises a bare `ErrorException`. If typed errors for GUI consumption are wanted, that is a better candidate than anything PCMM can offer here.

## `prepareBaseFile` on an unselected folder

`location == :rulesets_collection` was tested *before* the `ismissing(basename)` guard, so an unselected (`""`) rulesets folder skipped the `nothing` path every other location takes, went looking for `base_rulesets.csv` under a path with no folder component, and tripped an assertion inside PhysiCellXMLRules naming neither the location nor the folder.

To answer the question that came with the report: no, this does not belong in PhysiCellXMLRules. The assertion correctly describes that package's own contract; PCMM should not have made the call. Reordering the two branches is the whole fix.

## `configPath` motility paths

The three-token branch matched `token2 == "motility"` and sent **every** third token through `<options>`. But `<motility>` holds `speed`, `persistence_time` and `migration_bias` as direct children and reserves `<options>` for `enabled`, `use_2D`, `chemotaxis`:

```
before: configPath("default", "motility", "speed") -> .../motility/options/speed
after:  configPath("default", "motility", "speed") -> .../motility/speed
```

No results were ever silently wrong — variations are written through `setSimpleContent`, which calls `retrieveElement(...; required=true)` and throws. The element-creating `makeXMLPath` is not in the variation write path. The cost was a confusing "Element not found" for a spelling that reads as obviously correct.

The two-token spelling was always right, which is why this survived: the two spellings of the same parameter disagreed and only one was used. The regression test asserts their equality rather than a literal path.

Since `configPath`'s docstring invites guessing, the fix closes the class rather than the instance: unrecognized tokens in the two *closed* tag sets — motility and chemotaxis — are now rejected by name with the valid alternatives, matching every other `configPath` branch. `advanced_chemotaxis`, `phagocytosis`, `fusion`, `transformation`, `attack` and `custom` stay open-ended, because their third token is a substrate or cell type name PCMM cannot enumerate.

## ModelManager 0.9

Mechanical: compat bound, `packageName` removed (MM derives package identity from the module defining the simulator type), `getPackageVersion` → `getInstalledVersion`. `pcmmVersion()`'s docstring claimed "runtime version" while reporting the *installed* one; MM 0.9 separates those deliberately, so the docstring follows.

`rm_hpc_safe` changed contract — real removal attempted first, only residue staged, `:removed`/`:staged` returned, missing path with `force=false` still throws. `HPCTests` asserted the old "always stage into `data/.trash`" behaviour. Source needed no change; the tests did. The `:staged` branch is now deliberately uncovered: reaching it requires `rm` to fail on a path that still exists afterwards, which cannot be produced portably. The old tests only appeared to cover it because staging was unconditional.

## Making missing data audible

Four aggregation sites silently dropped replicates whose output is gone, which pruning makes reachable on ordinary data. Each now reports once:

```
[ Info: Excluding 1/3 replicates of monad 3 with no output on disk (deleted or pruned).
```

`@info` rather than `@warn` because pruning is a deliberate act — this reports a consequence, it does not scold. `maxlog=1` is load-bearing: calibration calls these once per monad across thousands of particles, and without it a pruned project buries the run in identical lines. Julia scopes `maxlog` per call site, so each site still reports independently.

A guard for the *other* silent case — a cell type present in some replicates of a monad but not others — was considered and rejected. It requires editing files under `data/`, which `best_practices.md` already forbids; the trust boundary sits at the data directory, and writing detection for a state only deliberate tampering can produce would assert we think it happens by accident.

## Docs

`docs/src/man/analyzing_output.md` described the plot recipes in prose and showed no figures at all. Three are committed under `docs/src/assets/`, generated by `docs/generate_figures.jl` so they can be remade when a recipe changes. Rendering them during `makedocs` would require the docs job to have a compiled PhysiCell and real output, which it does not.

## `Vector{QoI}` builders for the calibration summary statistics

ModelManager merged the QoI seam as #43 *during* this branch's review, so the migration `HANDOFF-QoI-unification.md` describes became possible and is included here. `endpointPopulationCountQoIs`, `endpointPopulationFractionQoIs` and `meanPopulationTimeSeriesQoIs` return the `Vector{QoI}` form of the three statistics, so the same quantity reaches sensitivity analysis and the post-processing sink rather than only `CalibrationProblem`. The monad-level functions are unchanged and remain the discover-everything path.

**One QoI per cell type, each named for its cell type.** `_asSummaryStatistic` wraps a reduced value as `Dict(q.name => value)`, so a single QoI named `"counts"` would hand `mseDistance` a `Dict("counts" => Dict("tumor" => …))` where it expects cell-type keys. That is also why `cell_types` is required: the vector is built before any simulation runs, so it cannot discover cell types from output.

**Three reducers, deliberately not shared.** `_reduceOverMonad` hands `reduce` every replicate's value including `missing`, so a default `mean` would return `missing` for any monad with a pruned replicate — reachable on ordinary data. Beyond that the three statistics disagree with each other:

| | cell type absent from a replicate | summation |
|---|---|---|
| counts | zero-filled | `mean` over a generator |
| fractions | zero-filled | `mean` over a materialised `Vector` |
| time series | **not** zero-filled — divides by replicates having it | `mean(array, dims=2)` |

A shared reducer would get one of them wrong silently. Each builder is asserted **equal** to its monad-level counterpart with `==`, not `isapprox`, both on an intact monad and on one with a replicate pruned — the case that actually distinguishes them.

The summation split is pinned with `fill(0.1, 1200)`. Above a blocksize of 1024 `sum` switches to pairwise summation, so the divergence follows from Julia's algorithm rather than from SIMD width and reproduces off the machine it was written on: `0.09999999999999988` against `0.09999999999999788`. An earlier draft searched random 24-element vectors instead, found none, and silently asserted nothing.

**One deliberate behavioural difference:** where *every* replicate is missing, `meanPopulationTimeSeries` raises a `KeyError` (its `cell_count` is left empty) while the builder returns `missing`, matching the other two.

## ModelManager moved during review

`main` advanced 13 commits while this was open, which broke two calibration tests: `runCalibration` now takes the method first, and generations became folders rather than flat `generation_<t>.csv` files (#39). Both fixed test-side. The generation assertion is now layout-agnostic rather than tracking the new filename — ModelManager reads both layouts, so the naming is its business, and persistence is already asserted through `posterior(result; generation=1)`.

More ModelManager work is expected before this release, so the drift needs re-checking rather than assuming it has settled.

## Not in this PR

- **PhysiPKPD inputs**, deferred to the CLAUDE.md to-do with its open questions written out: how a dosing schedule is described, where schedules live under `inputs/`, and how they participate in parameter variation.
- **`populationCountQoI` still returns a closure**, despite the name. It cannot simply be converted: it emits one sink column per cell type with the cell types read from the simulation's own output, while a `QoI` names its single column at construction. That needs either a rename or a ModelManager change letting one QoI contribute a `Dict` of `name => scalar`. Recorded in the to-do.


src/configuration.jl
```diff
@@ -87,17 +87,22 @@ makeXMLPath(x, xml_path::AbstractString) = makeXMLPath(x, [xml_path])
...
+unselected.
+
 Handles `:rulesets_collection` specially; all other locations use the generic MM default.
 """
 function prepareBaseFile(::PhysiCellSimulator, input_folder::InputFolder)
+    #! An unselected optional location has `basename === missing` and no folder to prepare, so
+    #! there is nothing to do for any location. This test must come first: `:rulesets_collection`
+    #! used to be checked ahead of it, so an unselected rulesets folder skipped this branch, went
+    #! looking for `base_rulesets.csv` under a path with no folder component, and tripped an
+    #! assertion inside PhysiCellXMLRules that named neither the location nor the folder.
```
drbergman: Is this short of this that if the `InputFolder` here was for the rules and it was missing, meaning it was not being used in this run, but the basename was somehow `base_rulesets.csv` that this would cause PhysiCellXMLRules to throw an error? If so, that seems quite convoluted and I don't see how we end up there. This feels more like an issue with how we parse the rules input in which if that holds an empty string (meaning not using rules), then we should build the InputFolder object accordingly, i.e. with basename set to missing. Here's MM's inner constructor for InputFolders:
```
    function InputFolder(location::Symbol, id::Int, folder::String)
        location_dict = inputsDict()[location]
        required = location_dict["required"]
        if isempty(folder)
            if required
                error("Folder for $location must be provided")
            end
            return new(location, id, folder, missing, required, false, "")
        end
        path_from_inputs = joinpath(location_dict["path_from_inputs"], folder)
        basename = location_dict["basename"]
        varied = folderIsVaried(location, folder)
        if basename isa Vector
            possible_files = [joinpath(locationPath(location, folder), x) for x in basename]
            basename_index = possible_files .|> isfile |> findfirst
            if isnothing(basename_index)
                error("Neither of $possible_files exist")
            end
            basename = basename[basename_index]
        end
        return new(location, id, folder, basename, required, varied, path_from_inputs)
    end
```
if `folder == ""`, then the basename is missing.

src/physicell_studio.jl
```diff
@@ -108,23 +108,25 @@ end
...
-            $(sprint(showerror, e))
-        """
-
-        return Base.IOError(msg, e.code)
+        #! `run` raises two different types here and they share no fields: `Base.IOError` (with
+        #! `.code`) when the executable cannot be spawned, and `ProcessFailedException` (with
+        #! `.procs`, no `.code`) when Studio runs and exits non-zero. Reading `.code`
+        #! unconditionally turned the second case into a `FieldError` from inside the handler.
+        #! Keep the cause intact instead and let the caller distinguish them.
+        (e isa InterruptException || !(e isa Exception)) && rethrow()
```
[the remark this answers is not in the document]
drbergman: `cleanUpStudioInputs` feels like it should be called from inside a `finally` block here? Everything seems to work already, so only make this change if it really improves things.

src/physicell_studio.jl
drbergman: shouldn't we handle the cleanup and everything inside here? or rather, I don't see why we just swallow this up inside the body of `runStudio` directly?

drbergman: but why not just fold this in to `runStudio`? Can't we do away with `executeStudio` being a separate function?

drbergman (commented):
looks good through ed8c565

PR #231, opened 2026-09-05
Ask SLURM for PhysiCell's thread count; track ModelManager 0.10

### LINKED ISSUE #234: Bare resume is impossible for any calibration built from PCMM's QoI builders
From the second review pass (2026-09-06). Related to #232 but narrower and independently actionable.

`docs/src/man/calibration.md` presents `resumeABC(Calibration(42))` as the way back into a run killed by a crash or an HPC time limit. For a problem built from PCMM's own builders it cannot work.

Every builder wraps lambdas for both halves of the `QoI` — `standard_qois.jl:216-217` and `:223` (via `_reduceKept` at `:184`), the same shape at `:250-267` and `:295-306`, and `post_processor_qois.jl:54`. ModelManager treats a lambda as unrestorable, so `problem.jld2` stores `nothing` for the summary statistic and bare resume raises "problem.jld2 contains only a partial manifest". PCMM's suite cannot see this: its only resume passes `problem=` explicitly (`CalibrationTests.jl:306`).

For the stated user base this is a real cost — a multi-day PhysiCell calibration hitting a wall-clock limit is the normal case, and the recovery the manual documents fails in a fresh session.

**Fix.** Make the builders callable structs rather than closures:

```julia
struct _EndpointCount <: Function
    cell_types::Union{Nothing,Vector{String}}
    include_dead::Bool
end
(c::_EndpointCount)(sim::Simulation) = ...
```

`nameof` of a callable struct is its type name, not a gensym, and JLD2 round-trips it as a type plus fields, so the QoI becomes restorable and the manifest complete. The same applies to the reducer. This is worth doing whether or not #232's contract change lands, and it is the mechanism that makes a *parameterised* measurement restorable at all — ModelManager's current definition of restorable ("a top-level named function") cannot express one.

**Test first:** add `resumeABC(result1.calibration; max_nr_populations=2)` with no `problem=`. It fails today.

Until it is fixed, `calibration.md` should show the `problem=` form first for PCMM users. (drbergman-lab/ModelManager.jl#56 tightens which functions count as restorable and makes an unreadable `problem.jld2` no longer block the `problem=` rescue, but it does not make a lambda restorable.)

drbergman:
From Claude: PhysiCell calls `omp_set_num_threads` with `parallel/omp_num_threads` from its config, so a job starts that many threads whatever SLURM allocated, and SLURM allocates one CPU unless asked. Every shipped sample config asks for 6 to 12. On a cluster each job therefore time-sliced its threads on one core, several times slower than a single-threaded run, with 100% CPU efficiency in `sacct` and nothing in any log to say why.

ModelManager 0.10 owns the `cpus-per-task` default and asks the backend through `simulationThreads(sim, simulation)`; PCMM implements it by reading that simulation's own `omp_num_threads` through the variation record, so a varied thread count is honoured. A user's own `cpus-per-task` replaces the default; a config whose element cannot be read falls back to 1 with one warning.

```julia
using PhysiCellModelManager          # on a cluster
mm_globals().sbatch_options["cpus-per-task"](simulation)   # -> 6 for the template config
setJobOptions(Dict("cpus-per-task" => 8))                  # your value wins
```

Tracking ModelManager 0.10 (compat `"0.10"`, version 0.5.0) also means adopting its contract, found by running this suite against ModelManager `main` with every 0.10 PR merged:

- A post-processor returning `nothing` is now refused ("say `missing`"), so `populationCountQoI` returns `missing` for a pruned snapshot, the manual's side-effect example returns `missing`, and `PrunerTests`' callback does too.
- `QoI`'s default `skip_missing=true` already does what `_reduceKept` did (drop the `missing` replicates, reduce an empty monad to `missing`), so that wrapper is gone from the three builders.
- `__init__` registers through `ModelManager.registerSimulator!` and `setNumberOfParallelSims` instead of assigning `mm_globals_ref` directly, which is internal again (its `@ref` was a `DocstringRefTests` failure).
- `march_flag` is set by `initializeModelManager` from ModelManager's HPC probe rather than by the `PhysiCellSimulator` constructor, which used to shell out to `which sbatch` during every dependent package's precompilation.
- `HPCTests` asserts `_SubmissionRefused` on the no-`sbatch` path (0.10 throws instead of recording a failed simulation) and the `simulationThreads` default.
- The QoI builders were closures over their keyword arguments, so `problem.jld2` stored `nothing` for them and a bare `resumeABC` of any PCMM calibration refused (#234). The keywords now travel in the QoI's `data` slot with named `compute`/`reduce` functions, which JLD2 restores by name; the resume test resumes without `problem=`. Closes #234.
- `post_processor_qois.jl` is folded into `standard_qois.jl`: it held `populationCountQoI` because that builder was written for the sink alone, and under 0.10 one `QoI` reaches every consumer, so the split described nothing. (Collapsing the two count builders themselves is #232.)
- `docs/src/lib/calibration.md` renders `samplePosterior` and `createTrial(::ABCResult, ::DataFrame)` from ModelManager #72; `checkdocs=:exports` fails the docs build without them.
- Manual: `mseDistance` compares the observed keys (an unresolvable observed key is an error, extra simulated components are ignored, one global mean over the differences computed); `distance` receives a `SummaryValues`; `populationCountQoI` is no longer described as sink-only, since the default reducer is a per-key mean.

Two manual items from the maintainer's to-do list, for 0.5.0: the sensitivity-analysis and calibration plot recipes are now shown as rendered figures (eight PNGs from `docs/generate_figures.jl`, which runs MOAT, RBD and a short ABC-SMC on the template project; the Sobol' bars are illustrative values, since a design that respects `ST ≥ S1` is 238 simulations for a picture whose job is to show the plot), and `prune_options` is documented on the post-processing page as the last step after a simulation, saying what each flag removes and what stops working afterwards.

Also: `requestHeaders` treats an empty `PCMM_PUBLIC_REPO_AUTH` as unset (an empty `Authorization` header made GitHub answer 401 to a request that succeeds anonymously), and the two comment-only placeholder source files are deleted.

CI cannot go green until ModelManager 0.10.0 is registered, since the compat bound points at it; the suite and docs build were run locally against ModelManager `main`.


docs/src/man/calibration.md
```diff
@@ -178,22 +178,22 @@ before.
...
+The keys of `observed_data` are the comparison: `mseDistance` resolves each one in the simulated summary, an observed key the summary statistic did not produce is an error rather than a silent zero, and an extra simulated component is ignored — a simulation is always known better than the data.
 
 ### Distance functions
 
-A distance function is any `(simulated, observed) → Float64`, where the argument types are whatever your `summary_statistic` returns and whatever you set `observed_data` to. [`mseDistance`](@ref) is the built-in option (dicts, vectors, scalars); a custom function can use any types:
+A distance function is any `(simulated, observed) → Float64`. `simulated` is a [`SummaryValues`](@ref) holding what each QoI's `reduce` returned; index it by cell type (`sim["cancer"]`) while only one QoI reports that key, or by `"<qoi name>.<key>"` when several do. `observed` is whatever you set `observed_data` to. [`mseDistance`](@ref) is the built-in option; a custom function can use any types:
```
drbergman: This is a terrible exposition. You don't "index it by cell type". You index it based on the keys your own QoIs name. We should name all 3 ways to index: By the key returned in your reduce function, by the `"<qoi name>.<key>"` convention, or the exact `(qoi_name, key)` Tuple.

docs/src/man/calibration.md
```diff
@@ -362,33 +362,25 @@ The original [`CalibrationProblem`](@ref) is loaded automatically from `problem.
...
     ```julia
     result = resumeABC(Calibration(42); problem = problem)
     ```
 
-    To avoid this requirement entirely, use named functions — either built-ins passed directly,
-    or functions defined at module level in your script:
+    **PCMM's QoI builders are closures** — `endpointPopulationCountQoI()` captures `cell_types` and
+    `include_dead` — so every calibration built from them needs `problem=` today (making them
+    restorable is issue #234). `mseDistance`, a top-level `my_stat`, and top-level `LatentVariation`
+    maps restore on their own:
```
drbergman: Unacceptable. One simple solution: include that in the `data` field of the QoI and tell users how to access. That's a little weird actually...because then they'll really need to know how that travels through the guts of MM. But we need something. It looks like #234 has a workable solution

drbergman: Does it require an `AbstractQoI` type in MM?

src/physicell_simulator.jl
```diff
@@ -40,13 +42,17 @@ end
...
+#! `initializeModelManager` sets `march_flag` from the `run_on_hpc` ModelManager has just probed,
+#! which is also one source of truth instead of two independent probes.
 function PhysiCellSimulator()
-    run_on_hpc = isRunningOnHPC()
     return PhysiCellSimulator(
         "",          # dir — set by initializeModelManager
         "g++",       # compiler — overridden by __init__ from ENV
         -1,          # current_version_id — set during DB init, refreshed at each compile
-        run_on_hpc ? "x86-64" : "native", # march_flag
+        "native",    # march_flag — set by initializeModelManager from ModelManager's HPC probe
```
drbergman: Why is this happening? MM should not be setting the march_flag for PCMM; MM should not know about the march flag for PCMM. I get that we want the check for running on HPC to happen elsewhere, but that value should be used here, right?

drbergman (commented):
Some comments

docs/src/man/calibration.md
drbergman: From Claude: Rewritten to name the three spellings `SummaryValues` resolves, in ModelManager's order: the key your own `reduce` returned (`sim["cancer"]`, while only one QoI reports it), the `"<qoi name>.<key>"` label the sink and sensitivity analysis also use, and the exact `(qoi_name, key)` tuple; a `Real`-valued QoI sits under its name alone. The `ts_dist` example below it indexes the same way. (commit 9623b3706)

docs/src/man/calibration.md
drbergman: From Claude: Agreed, and fixed in this PR rather than left to #234. No `AbstractQoI` needed, and nothing in ModelManager changes: the mechanism is the `data=` slot #70 added for exactly this. Each builder's keyword arguments now travel in `data`, and its `compute`/`reduce` are named top-level functions (`_endpointCountsOf`/`_meanEndpointCounts`, and so on), which JLD2 restores by name; `_isAnonymousFunction` deliberately ignores `data`, so the manifest is complete. The resume test now resumes bare (no `problem=`), and every builder is asserted restorable against ModelManager's own predicate. Callable structs would also work (`compute::Function` accepts any `<: Function`), but `data=` is what ModelManager's manual documents for a parameterised measurement, so users writing their own see one pattern. The admonition now says the builders restore on their own and points a user with parameters of their own at `data=`, not at how it travels. Closes #234. (commit 9623b3706)

src/physicell_simulator.jl
drbergman: From Claude: ModelManager does not set it and knows nothing about it; the comment was ambiguous because `initializeModelManager` is a ModelManager name that PCMM extends. It is PCMM's own `initializeModelManager(path_to_physicell, path_to_data)` method (`src/PhysiCellModelManager.jl`) that assigns `simulator().march_flag` from `mm_globals().run_on_hpc`, one line after ModelManager's init returns. The constructor cannot use the value: it runs in `__init__`, during precompilation, before any probe has happened, so the earliest moment the answer exists is right after ModelManager's init, and that is where PCMM reads it. Reworded the field docstring, the constructor comment and the test comment to say "PCMM's own method" and "ModelManager knows nothing about the flag". If you would rather the constructor never be mutated after the fact, the alternative is a `march_flag = nothing` default meaning "follow `run_on_hpc`" resolved at compile time; happy to do that instead, but it changes `setMarchFlag`'s tests and the `PCMM_MARCH` to-do design, so I left it. (commit 9623b3706)

src/analysis/standard_qois.jl
drbergman, line 1: Why do we have a distinction between post_processor_qois.jl and this file?

drbergman: From Claude: History, not design. `standard_qois.jl` dates from the modularization (2026-04) as "PhysiCell-specific calibration summary statistics"; `post_processor_qois.jl` was added with the `post_processor` hook (#217, 2026-07) because `populationCountQoI` was written for the sink alone: it had no reducer, and ModelManager's default `mean` could not combine keyed values, so its own header defined the split as "where the value goes". Under 0.10 one `QoI` reaches every consumer and the default reducer is a per-key mean, so that distinction describes nothing. Folded: `populationCountQoI` now lives at the bottom of `standard_qois.jl`, the file is gone, and the header says why it used to exist. Collapsing the two count builders themselves (`populationCountQoI` vs `endpointPopulationCountQoI`: zero-fill vs the default reducer, two column families for one quantity) is #232 and needs its own brief, so I left that alone. (commit 5e4831876)

drbergman:
From Claude: pushed 2e4afa6ea for the two manual items before 0.5.0. `docs/generate_figures.jl` now also runs MOAT(8), Sobolʼ(16), RBD(16) and a four-generation ABC-SMC on the template project (two parameters: a cycle phase duration and the apoptosis rate; final cell count) and saves eight PNGs, embedded in a new Plotting section of `sensitivity_analysis.md` and in `calibration.md`'s Visualization section. Pruning is documented on the post-processing page as the last of the post-completion steps, per your note, with the index pointing there; `PruneOptions`' docstring had lost `prune_xml`. Two cosmetic things the figures expose are ModelManager's, not PCMM's: the corner and transition recipes draw an empty top-right axis with 0–1 ticks, and ConvergenceSummary's 'max epsilon accepted' y-label clips in its third-height panel. Suite 730/730 and docs build clean against ModelManager main.

drbergman:
From Claude: f2cd8ed4d redraws the Sobol' figure. Two things were wrong with it: ModelManager's recipe drew ST translucent on top of S1, so the overlap was always a grey blend (fixed in drbergman-lab/ModelManager.jl#74, ST now behind an opaque S1), and the 16-sample design had estimated ST below S1, which no total-order index is. The design that respects that here is 238 simulations, so per your note the Sobol' bars are illustrative values through the real recipe, and the caption says so; MOAT, RBD and the calibration figures stay real. Docs build clean against the ModelManager fix branch.

PR #236, opened 2026-09-14
Report the cell type roster and IC setup failures honestly

### LINKED ISSUE #235: Two narrow analysis/diagnostics gaps found in the second review pass
From the second review pass (2026-09-06). Both are narrow and neither is a regression; filing so they are not lost.

## 1. `plotbycelltype` takes its cell-type roster from the trial's first simulation

The pruned-replicate fix (#228) corrected the denominator but not the roster. `population.jl:476-477` still does `simulation_id = simulationIDs(T) |> first; all_cell_types = cellTypeToNameDict(simulation_id) |> values |> collect`. If *that* simulation is the pruned one, PhysiCellOutput returns an empty `Dict` for a missing file, `processIncludeCellTypes(:all, …)` yields an empty vector, and the recipe draws a `(0, 1)` layout — an empty figure, no message. Whether it works depends on which replicate was pruned; the existing test prunes `last(sids)` (`PopulationTests.jl:85`), so it passes.

Fix: take the roster from the first simulation whose initial XML exists, or from the config. While there, `MonadPopulationTimeSeries` (`population.jl:257`) has the same skipped-replicate naming issue in its assertion message that `plotbycelltype` fixed at `:516` — it can name a replicate that was `continue`d.

## 2. An IC-cell/IC-ECM setup failure leaves no `output.err` to check

`prepareSimulationCommand` catches a `setUpICCell`/`setUpICECM` failure, prints a warning and returns `nothing` (`simulator_interface.jl:174-179`). ModelManager records the simulation as failed with `cmd === nothing`, so `postSimulationCleanup`'s `isnothing(cmd)` early return skips the `output.err` annotation — while `run`'s summary tells the user to "Check the output.err files for more information". On a sampling with a bad IC variation, the only diagnosis is a `println` interleaved with worker output.

Fix: write the cause into the simulation's `output.err` in those catch blocks, or have `postSimulationCleanup` annotate `output.err` whenever `!success` regardless of `cmd`. (The behaviour predates the `process` → `cmd` switch; that switch just rewrote this guard.)

drbergman:
From Claude: Two diagnostics that hid the diagnosis, from the second review pass (#235).

`plotbycelltype` read its cell-type roster from the trial's first simulation, so if that one replicate had been pruned the roster came back empty and the recipe drew an empty figure with nothing said. The roster now comes from the first simulation whose initial XML is still on disk, which any replicate can supply since they share a config; a trial with no readable output anywhere throws an `ArgumentError` naming it instead of drawing nothing. `MonadPopulationTimeSeries`'s time-grid assertion also now names the replicate it actually compared against rather than one it may have skipped.

An IC cell or IC ECM setup failure printed a warning and returned `nothing`, so ModelManager recorded the failure with no command and `postSimulationCleanup` never wrote `output.err`, while `run`'s summary told the user to check exactly that file. The cause is now written there before the `nothing`, the `println` is a `@warn`, and the `output.err` path has one definition shared with the cleanup step.

Both cases have tests: pruning the first replicate of a monad, and the existing failing-annulus IC variation asserting on the file. Closes #235.


src/analysis/population.jl
```diff
@@ -458,6 +464,31 @@ struct CellTypeInMonads
...
+function _trialCellTypeRoster(T::AbstractTrial)
+    for simulation_id in simulationIDs(T)
+        cell_type_to_name_dict = cellTypeToNameDict(simulation_id)
+        isempty(cell_type_to_name_dict) && continue
+        return values(cell_type_to_name_dict) |> collect
+    end
+    throw(ArgumentError("Cannot read the cell type roster of $(typeof(T)) $(T.id): none of its \
+                         $(simulationIDs(T) |> length) simulations has output on disk (deleted or pruned)."))
+end
+
```
drbergman: we should not read this from an `AbstractTrial`. It can only be read by an `AbstractSampling` since those are guaranteed to share config files.

src/analysis/population.jl
drbergman: we should probably shore up this logic as well. Insisting that we have an `AbstractSampling` and a better assertion statement/error

src/simulator_interface.jl
```diff
@@ -152,6 +152,48 @@ function setupSampling(::PhysiCellSimulator, S::AbstractSampling; force_recompil
...
+returns `nothing`, ModelManager records the failure with no `cmd`, and `postSimulationCleanup`
+early-returns on `isnothing(cmd)` without touching `output.err` — leaving `run`'s summary pointing
+at a file that does not exist. On a sampling where many simulations fail this way, the console is
+the only record and it is interleaved across workers; each simulation's own `output.err` is not.
+
+`@warn` rather than `println`: the message carries the simulation id into the logger's own stream
+rather than into whatever a worker happened to be writing mid-line, and a caller can silence or
+capture it like any other warning.
+"""
+function _recordSetupFailure(simulation::Simulation, ic_name::String, e)
```
drbergman: I think this name should be more specific to the IC part.

drbergman (commented):
Requires some review

src/simulator_interface.jl
```diff
@@ -185,7 +228,7 @@ function prepareSimulationCommand(simulation::Simulation)
         try
             append!(flags, ["-e", setUpICECM(simulation)])
         catch e
-            println("\nWARNING: Simulation $(simulation.id) failed to initialize the IC ECM file.\n\tCause: $e\n")
+            _recordSetupFailure(simulation, "IC ECM", e)
```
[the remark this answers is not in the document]
drbergman: From Claude: Added. `ICECMTests.jl` now asserts, for the failing elliptical-disc variation it already runs, that each simulation's `output.err` exists, names the IC ECM step, and carries a non-empty cause after the `---cause---` marker. (cbc7168be)

test/test-scripts/ICCellTests.jl
```diff
@@ -47,6 +47,22 @@ push!(discrete_variations, DiscreteVariation(xml_path, 300.0))
...
+#! `nothing`, ModelManager records the failure with no `cmd`, and `postSimulationCleanup`
+#! early-returns on `isnothing(cmd)` -- so nothing used to write `output.err`, while `run`'s closing
+#! summary told the user to go read it. The cause is now written there before the `nothing`.
+for simulation_id in simulationIDs(out_fail.trial)
+    path_to_err = PhysiCellModelManager._pathToSimulationErr(Simulation(simulation_id))
+    @test isfile(path_to_err)
+    err_contents = read(path_to_err, String)
+    @test occursin("IC cell", err_contents)
+    #! The cause itself, not just a header: the exception's own text, which is PhysiCellCellCreator's
+    #! to word, so match on the radii this variation set rather than on its sentence.
```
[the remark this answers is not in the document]
drbergman: From Claude: Fixed: "own wording". (cbc7168be)

src/analysis/population.jl
drbergman: From Claude: Agreed. The helper is `_samplingCellTypeRoster(::AbstractSampling)` and its docstring says why that is the widest type it can accept: a Sampling, Monad or Simulation shares one config across its simulations, a Trial does not. (cbc7168be)

src/analysis/population.jl
drbergman: From Claude: Done. The bare `@assert typeof(T) <: AbstractTrial` is now `T isa AbstractSampling || throw(ArgumentError(...))` with a message that says what the recipe draws (one panel per cell type of a single config), what it accepts (Sampling, Monad, Simulation, or `run`'s output on one), what it got, and to plot a Trial's samplings one at a time. The docstring's argument entry says the same, and PopulationTests asserts `plotbycelltype(Trial(1))` throws. (cbc7168be)

src/simulator_interface.jl
drbergman: From Claude: Renamed to `_recordICSetupFailure`, in the definition, both call sites and the two docstrings that mention it. (cbc7168be)

PR #237, opened 2026-09-14
One QoI builder per quantity, reduced by ModelManager's default mean

### LINKED ISSUE #232: Collapse the QoI builders once ModelManager's QoI contract is uniform
Companion to drbergman-lab/ModelManager.jl#52 (one QoI contract across the sink, sensitivity analysis and calibration). Nothing here should land before that does; recorded now so the PCMM side of the plan is not lost.

## What the review found (2026-09-05)

- `populationCountQoI` (sink only, no reducer, returns `nothing` on a missing snapshot, columns `population_count.<type>`) and `endpointPopulationCountQoI` (all consumers, bespoke reducer, returns `missing`, columns `endpoint_population_count.<type>`) measure the same thing at the default index. The split exists only because ModelManager's default `mean` cannot combine two `Dict`s. A user who stores with one and reads back (`stored=`) or analyses with the other gets two column families for one quantity.
- `meanPopulationTimeSeriesQoI` reaches calibration only: its `compute` returns a `SimulationPopulationTimeSeries` the sink cannot store and its `reduce` returns `Dict` of `Vector`s GSA refuses per component. The README, PRD, `calibration.md` and the source header explain the refusal as "a Vector is not spread by index"; the actual mechanism is that the `Dict` *is* spread and each component is then rejected for not being a `Real` (`sensitivity_analysis.md` has it right). README.md:91-92 are near-duplicate bullets about this.
- `_reduceKept`, `_excludedReplicates` and `_averageStatDicts` implement missing-replicate policy that belongs in ModelManager's `_reduceOverMonad`; the three builders carry three reducers that disagree on zero-fill because each mirrors its monad-level counterpart.
- `calibration.md` (~line 366) says closures that capture variables are detected and stored as `nothing`, and (~line 386) that the built-in builders return a "named, serializable measurement". Every builder uses lambdas, so every PCMM calibration needs `problem=` on resume; ModelManager's `_isAnonymousFunction` fix (its next release) changes what is detected but not this.

## Plan (after ModelManager ships the contract)

1. Delete `_reduceKept` and the three bespoke reducers; the builders become `compute`-only with ModelManager's keyed mean.
2. Keep one count builder: `populationCountQoI(; index=:final, cell_types, include_dead)`, returning `missing` (not `nothing`) for a missing snapshot; `endpointPopulationCountQoI` becomes a one-release deprecated alias or is removed. Same collapse for fractions (`populationFractionQoI`).
3. Replace `meanPopulationTimeSeriesQoI` with `populationTimeSeriesQoI(; times, cell_types, include_dead)` whose `compute` returns a flat `Dict("<type>@<t>" => count)` for the requested snapshot times (error if a time is not on the output grid). `times=` is required: a 30-day run saved hourly × 3 cell types would exceed SQLite's default 2000 columns in the sink.
4. The monad-level `endpointPopulationCounts` / `endpointPopulationFractions` / `meanPopulationTimeSeries` stay as analysis helpers, or are re-expressed as "evaluate this QoI on this monad" so they cannot drift from the builders.
5. Docs: `observed_data` keyed by label (`population_count.tumor`), the "which builder reaches which consumer" matrix deleted from README/PRD/`calibration.md`, the closure/serializable claims corrected. Consider a callable-struct idiom for the builders so they are restorable across sessions.

drbergman:
From Claude: Resolves #232 per the maintainer's decisions of 2026-09-14, in three commits.

**One reducer.** No builder defines a `reduce`; every builder is averaged by ModelManager's default per-key mean through the two-argument form `data` selects (needs drbergman-lab/ModelManager.jl#76, now merged). The bespoke reducers existed to reconcile a replicate lacking a cell type, which cannot happen: `populationCount` keys every cell type the model declares, and replicates of a monad share a config. `meanPopulationTimeSeriesQoI`'s `compute` returns the per-cell-type series `Dict` so the default mean averages it elementwise.

**One builder per quantity, each taking an `index`.** `populationCountQoI(; index=:final, cell_types, include_dead)` and the new `populationFractionQoI` with the same signature, sharing one snapshot read; sink columns and labels are `population_count.<cell_type>` and `population_fraction.<cell_type>`. `endpointPopulationCountQoI` and `endpointPopulationFractionQoI` are removed rather than aliased, since `:final` is the default.

**No monad-level statistics.** `endpointPopulationCounts`, `endpointPopulationFractions` and `meanPopulationTimeSeries` took a monad ID and duplicated the builders under the default reducer; removed with `_averageStatDicts`. `finalPopulationCount(::Monad)` and `MonadPopulationTimeSeries` remain for direct monad analysis. The builder-vs-monad-level tests are replaced by direct value tests: the reduced count equals the mean of the replicates' `finalPopulationCount`, fractions sum to 1 per replicate and after reduction, and the time-series builder equals the elementwise mean of the replicates' series.

Breaking, as 0.5.0 is. Suite 726/726 and docs build clean against ModelManager `main`. Closes #232.


docs/src/man/calibration.md
```diff
@@ -573,11 +573,13 @@ without naming them in advance, labelled `endpoint_population_count.<cell_type>`
...
-    mean applies wherever it is reduced, and that default refuses a monad whose replicates report
-    different cell types. [`endpointPopulationCountQoI`](@ref) measures the same thing at the final
-    snapshot and zero-fills a cell type a replicate lacks.
+!!! note "Two builders, one reducer"
+    [`populationCountQoI`](@ref) and [`endpointPopulationCountQoI`](@ref) measure the same quantity
+    at the final snapshot, under two names and two families of sink columns. Neither defines a
+    `reduce`, nor does any other builder on this page: a monad's replicates are averaged by
+    ModelManager's default per-key mean throughout. That default asks the replicates to report the
+    same cell types, which they always do — the keys are the model's declared cell-type roster, and
+    replicates of a monad share a config.
```
drbergman: Why even call this out?

drbergman:
From Claude: two more commits on top per today's decisions: e0e31692 drops `endpointPopulationCountQoI` for `populationCountQoI`; e6469224 renames the fraction builder to `populationFractionQoI(; index)` sharing the snapshot read with the count builder, and removes the three monad-level statistics with their tests replaced by direct value checks. Body and title rewritten. 726/726, docs clean.

PR #238, opened 2026-09-15
Tiered manual with a home for every public name

drbergman:
From Claude:

Of the 300 names a user can reach through `using PhysiCellModelManager` (exported or `public`, PCMM's own or re-exported from ModelManager), 191 had no manual page; `samplePosterior` was the example that surfaced it. Each name is now classed end-user (113) or developer (187). User names get ordinary prose on task pages; developer names sit in `!!! tierdev` blocks beside the code they concern, or on three new Dev-tier pages (architecture, simulator interface, utilities). A sidebar Detail selector (Code / Brief / Full / Dev / Journal) hides or reveals those blocks client-side, so the source and the built HTML always carry every tier.

Three new user pages (running simulations, the trial hierarchy, managing a project); calibration, sensitivity analysis, analysing output, post-processing and tagging are tiered; the calibration page now teaches `run(method, problem)` and demotes `runABC`/`runCalibration` to the Dev tier. A new test fails when any public name has no page. `AGENTS.md` and `CONTRIBUTING.md` replace the five-line developer guide.

Two things outside `docs/`: `simulationRuntime`'s docstring was detached by a blank line and its doctest referenced an undefined variable (now a plain example), and `CLAUDE.md` gains a one-line pointer to `AGENTS.md`.

The push preview is the review surface — read the finished pages at each tier rather than the diff. Worth a look while reviewing: whether the design-method types (`LHSVariation` etc.) should be promoted back to user prose on the varying-parameters page, and whether the v0.3.3 column-rename note on the post-processing page should stay visible to everyone rather than Dev only.


docs/src/dev/architecture.md
```diff
@@ -0,0 +1,250 @@
...
+| `src/analysis/standard_qois.jl` | Ready-made [`QoI`](@ref) builders: [`populationCountQoI`](@ref), [`populationFractionQoI`](@ref), [`meanPopulationTimeSeriesQoI`](@ref). |
+
+## From createTrial to a finished simulation
+
+One call, start to finish. Everything named as a hook is defined in ModelManager and implemented
+for PhysiCell in this package; the [Simulator interface](@ref simulator_interface_dev) page gives
+the contract for each.
+
+```julia
+run(createTrial(inputs, variations; n_replicates=3))
```
drbergman, line 73: why not also highlight here that `run` can act on the exact arguments being passed in to `createTrial` here?

docs/src/dev/simulator_interface.md
```diff
@@ -0,0 +1,140 @@
+# [Simulator interface](@id simulator_interface_dev)
+
+The contract PhysiCellModelManager.jl fulfils for ModelManager: every hook, what ModelManager
```
drbergman: "fulfil" is a British spelling. no British spellings!

docs/src/dev/simulator_interface.md
```diff
@@ -0,0 +1,140 @@
...
+ModelManager owns the database; the backend supplies the names and the simulator-specific schema
+fragments. These are pure functions of the backend — no state, no I/O — which is what lets
+ModelManager build queries before a project is open.
+
+| Hook | ModelManager expects | PCMM |
+|---|---|---|
+| [`centralDBFileName`](@ref) | The filename (not path) of the central SQLite database. | `"vct.db"` when that legacy pre-PCMM file exists in the data directory, otherwise `"pcmm.db"` — `src/globals.jl`. |
+| [`dbVersionTableName`](@ref ModelManager.dbVersionTableName) | The table that records the package version. | `"pcmm_version"` — `src/simulator_interface.jl`. |
+| [`simulatorVersionTableName`](@ref ModelManager.simulatorVersionTableName) | The simulator version table. | `"physicell_versions"` — `src/simulator_interface.jl`. |
+| [`simulatorVersionIDName`](@ref ModelManager.simulatorVersionIDName) | The FK column naming that table from `simulations`, `monads` and `samplings`. | `"physicell_version_id"` — `src/simulator_interface.jl`. |
```
drbergman: no idea what `FK column` means

docs/src/dev/utilities.md
```diff
@@ -0,0 +1,128 @@
...
+end
+
+tableColumns("notes")   # ["note_id", "simulation_id", "body"]
+```
+
+## XML
+
+PhysiCell's inputs are XML, so most of PCMM's configuration code is element lookup. Every helper
+takes the path as a vector of tag names, with `tag:attribute:value` to select among siblings by
+attribute and `tag:child_tag:content` to select by a child's content — [`getChildByAttribute`](@ref ModelManager.getChildByAttribute)
```
drbergman: isn't this `tag::child_tag:content`

docs/src/man/analyzing_output.md
```diff
@@ -1,100 +1,231 @@
...
+using Plots
+plot(sampling)                                       #! one panel per monad
+plot(sampling; include_dead=true)                    #! count dead cells too
+plot(sampling; include_cell_type_names="cancer")     #! restrict to one cell type
+plot(sampling; exclude_cell_type_names="cancer")     #! or drop one
+plot(sampling; time_unit=:h)                         #! x-axis in hours
+```
 
-One panel per `Monad`, with each cell type a series and the shaded band its standard deviation across replicates:
+One panel per `Monad`, with each cell type a series and the shaded band its standard deviation
```
drbergman: need to explain that the title defaults to the parameter values that define each Monad relative to the others in the Sampling

docs/src/man/analyzing_output.md
```diff
@@ -1,100 +1,231 @@
...
 ```julia
-data = cellDataSequence(1, "position")
+sequence = PhysiCellSequence(1; include_cells=true) #! every snapshot of simulation 1
 ```
-Then one can access the positions of the cell with ID 78 by
+
+### Loading data into a snapshot or sequence
+
+!!! tiergloss
+    A snapshot or sequence built without the `include_*` keywords carries no data. Four functions
```
drbergman: This should be exemplified below

docs/src/man/analyzing_output.md
```diff
@@ -115,144 +249,172 @@ plot(ests.time, ests["cd8"]["IFNg"])
...
+!!! tiergloss
+    `pcf` accepts a `PhysiCellSnapshot`, a `PhysiCellSequence`, or a `Simulation`. An `Integer`
+    first argument is treated as a simulation ID; follow it with an index (an `Integer`, or
+    `:initial` / `:final`) to compute at one snapshot rather than over the whole simulation. Next
+    comes the center cell type (a `String` or `Vector{String}`), then optionally the target cell
+    type; omit the target and the centers are used, giving a non-cross PCF. The center and target
+    sets must be identical or disjoint. Call it as `PhysiCellModelManager.pcf`, or as plain `pcf`
+    once `using PairCorrelationFunction` has been called.
 
 ### Keyword arguments
```
drbergman: why does PCF alone get a keyword arg section? These are in many of the plotting functions.

docs/src/index.md
```diff
@@ -12,6 +12,8 @@ CurrentModule = PhysiCellModelManager
 
 New here? Start with [Installation](@ref installation_man), then [Your first project](@ref getting_started_man).
 
+The **Detail** selector in the sidebar sets how much each page explains — **Code** shows only the calls, **Brief** adds a line per step, **Full** adds the reasoning, **Dev** adds what someone extending PCMM needs and reveals the Developers pages, **Journal** adds dated design notes. The choice is remembered across pages.
```
drbergman: could we give this a header so users can more readily pick this out? and maybe have this be a table rather than a paragraph?

docs/src/man/varying_parameters.md
```diff
@@ -37,7 +37,20 @@ variationName(dv)
...
 
-If `name` is omitted, PhysiCellModelManager.jl assigns a default based on the target/location naming conventions used by [`shortVariationName`](@ref PhysiCellModelManager.shortVariationName).
+!!! tierdev
+    **Reading a name back.** [`variationName`](@ref) returns the string a variation reports under.
+    If `name` was omitted at construction, that string is the convention-based default
+    `shortVariationName(location, columnName(target))`.
+    [`shortVariationName`](@ref PhysiCellModelManager.shortVariationName) dispatches on the
+    simulator — ModelManager's fallback returns the column name unchanged, and
+    PhysiCellModelManager.jl extends it to produce the readable `"default: apoptosis death rate"`
+    form. Extend it for a new simulator rather than renaming columns.
```
drbergman: Why are we giving suggestions here in PCMM to somebody building on MM?

docs/src/man/varying_parameters.md
drbergman, line 1: I think we can furtehr condense this in the "code" view

docs/src/man/varying_parameters.md
```diff
@@ -61,10 +74,67 @@ d = Uniform(0, 0.001)
...
 
 ```julia
 dv = DistributedVariation(xml_path, d; name="apoptosis rate")
 ```
 
-These variations are useful for doing [Sensitivity analysis](@ref sensitivity_analysis_man).
\ No newline at end of file
+These variations are useful for doing [Sensitivity analysis](@ref sensitivity_analysis_man).
+
+## Sampling the variation space
```
drbergman: We need less here for "code" (just the calls), more for Brief (what each of the methods does briefly) and Full (a full treatment of each linking the index for the docstrings)

docs/src/man/varying_parameters.md
```diff
@@ -61,10 +74,67 @@ d = Uniform(0, 0.001)
...
+    `variation_matrix`, the IDs re-sorted into the layout RBD-FAST spectral analysis consumes. Call
+    `addVariations` directly only when you want the IDs without building a trial around them.
+
+    **Domain bounds in bulk.** [`domainVariations`](@ref) is a PhysiCellModelManager.jl helper that turns a
+    named tuple of domain boundaries into the corresponding `DiscreteVariation`s, so a domain-size
+    sweep does not need six hand-written `configPath` calls. Keys are matched loosely — each must
+    contain `min` or `max` and one of `x`, `y`, `z` — so `x_min`, `xmax`, and `min_y` all work, and
+    you need not give all three dimensions. Values may be scalars or vectors; with `covary=true` it
+    returns a single `CoVariation` stepping the boundaries together (every multi-valued boundary must
+    then have the same number of values) instead of a full factorial over them.
```
drbergman: this is out of place. this subsection on this page should just be about the GridVariation, LHSVariation, etc.

docs/src/man/latent_variations.md
```diff
@@ -11,7 +11,14 @@ To construct one, provide:
...
 
 ## Latent Parameter Names
-Optionally name the latent parameters; the names appear in the `LatentVariation` display, which helps when reading sensitivity-analysis or optimization results. If omitted, names default from the target parameters and their index (the target portion follows PhysiCellModelManager.jl short variation naming). See [`defaultLatentParameterNames`](@ref PhysiCellModelManager.ModelManager.defaultLatentParameterNames).
+Optionally name the latent parameters; the names appear in the `LatentVariation` display, which helps when reading sensitivity-analysis or optimization results. If omitted, names default from the target parameters and their index (the target portion follows PhysiCellModelManager.jl short variation naming).
+
+!!! tierdev
+    [`defaultLatentParameterNames`](@ref PhysiCellModelManager.ModelManager.defaultLatentParameterNames)
+    builds those fallback names: `"<target_1> | <target_2> | … | lp#<i>"`, joining every target's
```
drbergman: this is something a user would be interested in knowing since they are likely to see this. put it in "Full"

docs/src/man/latent_variations.md
drbergman, line 1: "Code" should be much more compact

docs/src/man/intracellular_inputs.md
drbergman, line 1: "Code" should be much more compact

docs/src/man/running_simulations.md
```diff
@@ -0,0 +1,119 @@
...
+    [`useHPC`](@ref) pins simulations to SLURM submission for the session;
+    [`setJobOptions`](@ref) merges entries into the `sbatch` options dictionary, each becoming a
+    `--key=value` flag. A value may be a function of the [`Simulation`](@ref) about to be
+    submitted, so an option can follow a varied parameter.
+
+```julia
+useHPC()                                  # useHPC(false) forces local runs
+setJobOptions(Dict("time" => "02:00:00",
+                   "mem" => "8G",
+                   "partition" => "shared"))
```
drbergman, line 95: somewhere around here (in "Brief" or "Full") we need to mention the default number of CPUs being tied automatically by PCMM to the number of threads in the config file, warning the user not to touch that

docs/src/man/running_simulations.md
```diff
@@ -0,0 +1,119 @@
...
+    unless `MODELMANAGER_HPC_DONE_DIR` says otherwise, and it is fixed for the session.
+
+!!! tiergloss
+    A submitted job reports its exit code by writing a sentinel file to a shared directory, which
+    the submitting worker polls. [`setHPCCompletionOptions`](@ref) adjusts the timings of that
+    protocol, which are the fields of [`HPCCompletionOptions`](@ref ModelManager.HPCCompletionOptions):
+    `submit_retry_period` (how long a transiently refused submission is retried),
+    `poll_interval` (how often a worker checks for its own sentinel), `reap_interval` (how long
+    one `squeue` answer is shared among waiting workers), and `grace_period` (how long a job may
+    be missing from the queue with no sentinel before it is declared failed).
```
drbergman: I don't think I agree with this being "Brief". I think this is "Full"

docs/src/man/trials.md
```diff
@@ -0,0 +1,109 @@
...
+|---|---|---|
+| [`Simulation`](@ref) | One run of the model | one parameter set, one seed |
+| [`Monad`](@ref) | Replicates of one parameter set | simulations |
+| [`Sampling`](@ref) | Monads on one [`InputFolders`](@ref), differing in parameters | monads |
+| [`Trial`](@ref) | Samplings, which may use different [`InputFolders`](@ref) | samplings |
+
+!!! tierwhy
+    You do not choose the class; you describe the runs and the class follows. That is what lets a
+    script grow a `DiscreteVariation` from one value to three without any other edit — the same
+    call that returned a `Simulation` now returns a `Sampling`, and [`run`](@ref) takes either.
```
drbergman: I think positioning these `tierwhy` below the `tiergloss` makes more sense, a descent into detail

docs/src/man/trials.md
```diff
@@ -0,0 +1,109 @@
...
+!!! tiergloss
+    [`createTrial`](@ref) returns the narrowest class that fits what you asked for: no variations
+    and one replicate gives a `Simulation`, several replicates of one parameter set a `Monad`, and
+    a variation with more than one value a `Sampling`. Pass a vector of already-built trials to
+    bundle them into a `Trial`.
+
+```julia
+inputs = InputFolders("0_template", "0_template")
+
+createTrial(inputs)                                                   # Simulation
```
drbergman, line 34: at the Full level, we should note that a Monad is automatically every time a Simulation is created. There are not Simulation's without corresponding Monad objects.

docs/src/man/trials.md
```diff
@@ -0,0 +1,109 @@
...
+    for a one-off; it issues one query each, which is what matters at campaign scale.
+    `printSimulationIDs` takes an `IO` first — `printSimulationIDs(stdout, trial)` — and prints
+    the hierarchy as an indented tree.
+
+## Read what a trial contains
+
+!!! tierwhy
+    Columns that are constant across every row are dropped by default, so the table shows only
+    what actually varied. Pass `remove_constants = false` when you want the full parameterization
+    — for instance when the table is going into a paper's supplement rather than onto your screen.
```
drbergman: I think the "for instance" is to just see all the parameters in the database.

docs/src/man/trials.md
```diff
@@ -0,0 +1,109 @@
...
+    Add `tags = true` to group rows by what a run was for — see
```
drbergman: does this sort by tags if they're included? Or do they need to pass in the sort list?

docs/src/man/sensitivity_analysis.md
```diff
@@ -1,138 +1,185 @@
...
-A sensitivity analysis takes the same inputs as a sampling:
-- `inputs::InputFolders` — the `data/inputs/` folders defining your model.
-- `evs::Vector{<:ElementaryVariation}` — the parameters to analyze and their ranges/distributions.
 
-Unlike most trials, these are usually [`DistributedVariation`](@ref)s, so a continuum of values can be tested. Use the convenience constructors [`UniformDistributedVariation`](@ref) and [`NormalDistributedVariation`](@ref), or any `d::Distribution` directly:
+!!! tierwhy
+    A sensitivity analysis takes the same inputs as a sampling: an `inputs::InputFolders` naming the
+    `data/inputs/` folders that define your model, and an `evs::Vector{<:ElementaryVariation}` of
+    the parameters to analyze with their ranges or distributions. Unlike most trials these are
+    usually [`DistributedVariation`](@ref)s, so a continuum of values can be tested.
```
drbergman: Do not make statements like this. The "unlike most trials". These are, in fact, often distributions.

docs/src/man/sensitivity_analysis.md
```diff
@@ -1,138 +1,185 @@
...
+NormalDistributedVariation(xml_path, 1e-3, 1e-4; lb=0)          # (mean, std), truncated at lb
 ```
 
-[`CoVariation`](@ref)s draw all member parameters from the same CDF value; pass `flip` to negatively correlate some of them. For more complex relationships, use [LatentVariations](@ref latent_variations_man) to transform latent variables into the parameters of interest.
+### Sensitivity functions
 
-All variation types accept `name=...`, used in the scheme DataFrame/CSV headers. Inspect the effective name with [`variationName`](@ref).
+!!! tierwhy
+    A sensitivity function takes a single argument, a `Simulation`. Declare it `::Simulation`:
+    an unannotated function cannot be told apart from one written to take a simulation ID (`Int`).
```
drbergman: This is an example of you explaining the journal of what we did and not what a user needs to know. Also, do we still assert that it must have an explicit signature like this? Or can it accept anything? I think all we need to do here is tell users they must pass in a function that accepts `Simulation`

docs/src/man/sensitivity_analysis.md
```diff
@@ -1,138 +1,185 @@
...
-### Sensitivity functions
-At the time of starting the sensitivity analysis, you can include any number of sensitivity functions to compute.
-They must take a single argument, a `Simulation`; declare it `::Simulation`, since an unannotated function cannot be told apart from one written to take a simulation ID (`Int`). A bare function's per-replicate values must average to a `Real`; passed as a [`QoI`](@ref ModelManager.QoI), `reduce` may instead return a `Dict`/`NamedTuple` of `Real`s, which spreads — see below.
-For example, `finalPopulationCount` returns a dictionary of the final population counts of each cell type from a `Simulation`.
-So, if you want to know the sensitivity of the final population count of cell type "cancer", you could define a function like:
 ```julia
 f(sim::Simulation) = finalPopulationCount(sim)["cancer"]
 ```
 
 ### [One measurement, one analysis per cell type](@id gsa_keyed_qoi)
```
drbergman: What is this section even about? I really think it should be removed.

docs/src/man/sensitivity_analysis.md
```diff
@@ -1,138 +1,185 @@
...
-Naming a cell type up front means one function per cell type. Pass a [`QoI`](@ref ModelManager.QoI)
-whose `reduce` returns a `Dict` instead, and each key becomes its own sensitivity analysis labelled
-`"<qoi name>.<key>"`. The ready-made builders do this already:
+
+!!! tierwhy
+    Naming a cell type up front means one function per cell type. Pass a
```
drbergman: What does this even mean? "Naming a cell type up front"? This feels like nonsense

docs/src/man/sensitivity_analysis.md
```diff
@@ -1,138 +1,185 @@
...
+    [`populationCountQoI`](@ref).
+
+!!! tierwhy
+    **Two shapes that are not spread.** Two separate rules. A bare `Vector` return is **not** spread
+    by index — components are keyed, for now, so that two parameter sets can be checked for the same
+    components by name. Separately, every spread component must itself be a `Real`. It is the second
+    that rules out [`meanPopulationTimeSeriesQoI`](@ref): its `Dict` *is* spread, and each value is
+    then rejected for being a time series rather than a number. Reduce a series to a scalar to ask a
+    sensitivity question about it. No builder defines a `reduce` of its own, so ModelManager's
+    default per-key mean averages the replicates of all of them.
```
drbergman: I also think this is not the right place to be explaining QoIs and naming conventions

docs/src/man/sensitivity_analysis.md
```diff
@@ -143,49 +190,65 @@ evs = [NormalDistributedVariation(configPath("cancer", "apoptosis", "rate"), 1e-
...
+    `calculateGSA!` method that joins them.
+
 ## Post-processing
-The object `sensitivity_sampling` is of type [`GSASampling`](@ref PhysiCellModelManager.ModelManager.GSASampling), meaning you can use [`PhysiCellModelManager.calculateGSA!`](@ref) to compute sensitivity analyses.
+
+!!! tierwhy
+    Results accumulate, and a measurement whose name is already present is **skipped** — so reusing
+    the name `f` from the run above would do nothing at all, silently. Pass `recompute=true` to
+    force re-evaluation; it has to be explicit, because redefining a function's body leaves it
+    indistinguishable from the one already evaluated. Results are stored in a `Dict` keyed by the
```
drbergman: "Pass `recompute=true` to force re-evaluation rather than using the post-processing database that has stored QoIs." Something like that. We need to make sure that we either reference the database here or link to its description. This "why" only makes sense in the context of that database.

docs/src/man/calibration.md
```diff
@@ -302,114 +443,101 @@ method = ABCSMC(perturbation_kernel = LocalNNKernel(k = 15))
...
 # Local covariance based on k nearest neighbours
 method = ABCSMC(perturbation_kernel = LocalNNCovKernel(k = 15))
-```
-
-Both `GaussianKernel` and `ComponentwiseKernel` accept an optional `scale` multiplier (default `2.0`) applied to the weighted (co)variance:
 
-```julia
+# GaussianKernel and ComponentwiseKernel take an optional `scale` on the weighted (co)variance
 GaussianKernel(1.0)          # scale = 1 × covariance
 GaussianKernel([1.0, 2.0])   # per-generation scale vector
```
drbergman: What happens when we go beyond generation `t=2`?

docs/src/man/calibration.md
```diff
@@ -302,114 +443,101 @@ method = ABCSMC(perturbation_kernel = LocalNNKernel(k = 15))
...
+result = resumeABC(calibration)          # the ABC-specific alias, same arguments
+result = run(calibration)                # and the same thing through `run`
 
 # Patch one setting — e.g. allow more generations than the original run
-result = resumeABC(calibration; max_nr_populations = 20)
+result = resumeCalibration(calibration; max_nr_populations = 20)
 
 # If the original problem used anonymous functions (not serializable), re-supply it:
-result = resumeABC(calibration; problem = problem)
+result = resumeCalibration(calibration; problem = problem)
```
drbergman: Let's pick one call and stick with it and then show at the end the alternatives. If we can get this done through `run`, let's just use that. If we need `resumeCalibration` for some reason, we should revisit and see if we can actually go through run. Any other grid sampling or sensitivity can be resumed via a call to run.

docs/src/man/calibration.md
```diff
@@ -302,114 +443,101 @@ method = ABCSMC(perturbation_kernel = LocalNNKernel(k = 15))
...
+lv = LatentVariation(
+    [Uniform(0.0, 1.0)],
+    [configPath("cancer", "apoptosis", "rate")],
+    [apoptosis_map];
+    inverse_maps = [apoptosis_inv],
+)
+
+# Anonymous: problem.jld2 will be incomplete, so keep `problem` to re-supply on resume
+problem = CalibrationProblem(ref, params, observed,
+    sim -> finalPopulationCount(sim),   # a lambda cannot be restored by name
```
drbergman: Let's not show a poor way to do this. Make up a better function (even `sim -> sim.id` would work; just not something that is already named)

docs/src/man/calibration.md
```diff
@@ -302,114 +443,101 @@ method = ABCSMC(perturbation_kernel = LocalNNKernel(k = 15))
...
+    mseDistance)
+result = resumeCalibration(Calibration(42); problem = problem)
+```
 
 ## Inspecting results
 
 ### Posterior samples
```
drbergman, line 549: We have to cover `samplePosterior` here.

docs/src/man/calibration.md
```diff
@@ -452,10 +597,15 @@ plot(ConvergenceSummary(result))
...
+plot(Calibration(42), :ridgeline)
 ```
 
-On a two-parameter example — a cycle phase duration and an apoptosis rate, calibrated to one
-simulation's final cell count on the template project — the four plots look like this:
+!!! tiergloss
+    The figures below come from a two-parameter example: a cycle phase duration and an apoptosis
+    rate, calibrated to one simulation's final cell count on the template project.
 
 ![Corner plot of the final-generation posterior](../assets/calibration_corner.png)
```
drbergman, line 610: feels like this plot should have empty axes on the upper triangle, not a set of axes drawn with nothing plotted on it.

docs/src/man/calibration.md
```diff
@@ -452,10 +597,15 @@ plot(ConvergenceSummary(result))
 # Generation-transition plot: gen-t posterior + gen-(t+1) proposals (accepted=green, rejected=red)
 plot(result, :transition)                  # last complete transition
 plot(result, :transition; generation = 2)  # specific transition t → t+1
+
+# From disk, with no in-memory result
+plot(Calibration(42))
+plot(Calibration(42), :ridgeline)
 ```
```
drbergman, line 608: we should have these calls next to the plots they represent rather than force users to identify which is which

docs/src/man/calibration.md
```diff
@@ -251,37 +361,68 @@ method = ABCSMC(
 )
 ```
 
-Generation `t` uses `epsilon_schedule[t-1]`; the schedule takes precedence over `epsilon_quantile`.
-`min_acceptance_rate` is useful as a safety stop alongside a schedule.
-
 ### Simulation bank and CDF-grid snapping (`cdf_grid_k`)
```
drbergman: We need a "Brief" section here to give some context

docs/src/man/calibration.md
```diff
@@ -302,114 +443,101 @@ method = ABCSMC(perturbation_kernel = LocalNNKernel(k = 15))
...
+    mseDistance)
+result = resumeCalibration(Calibration(42); problem = problem)
+```
 
 ## Inspecting results
 
 ### Posterior samples
 
+!!! tiergloss
+    [`posterior`](@ref) returns `(df, weights)`: a `DataFrame` with one column per calibrated
```
drbergman: is it really returning a tuple and not the df with the weights included as a column? Shouldn't MM return just the dataframe?

docs/src/man/calibration.md
drbergman: ok, I see now why they're split. we need a link here to the `samplePosterior` section below

docs/src/man/calibration.md
```diff
@@ -495,89 +711,102 @@ data/outputs/calibrations/1/
...
+    replicates. Use it when `observed_data` is a time series rather than a single endpoint value.
+
 ```julia
 meanPopulationTimeSeriesQoI(; cell_types=nothing, include_dead=false)
 ```
 
-Each cell type's count over time, on the replicate's own grid, averaged elementwise across
-replicates. Use it when `observed_data` is a time series rather than a single endpoint value.
-
 ### [QoI form](@id qoi_form_ss)
```
drbergman: I don't understand the title of this section. The above ### sections are also the QoI form.

docs/src/man/calibration.md
```diff
@@ -1,14 +1,28 @@
...
+    inference method. It refines a population of parameter samples (particles) over successive
+    generations, keeping only those whose simulated output falls within a shrinking tolerance
+    (epsilon) of the observed data. An agent-based model has no tractable likelihood, so the method
+    that applies is the one that only ever needs to *run* the model.
+
+    **What comes from where.** The implementation is native Julia — no Python or conda environment
+    is required. The algorithm lives in ModelManager; PhysiCellModelManager.jl contributes the
+    PhysiCell-specific summary statistics ([`populationCountQoI`](@ref),
+    [`populationFractionQoI`](@ref), [`meanPopulationTimeSeriesQoI`](@ref)). One
+    `using PhysiCellModelManager` brings both.
```
drbergman: this feels like it belongs either for devs to know or only just in the journal

docs/src/man/calibration.md
```diff
@@ -40,33 +54,37 @@ problem = CalibrationProblem(
...
 ## Defining the calibration problem
 
-[`CalibrationProblem`](@ref) bundles everything the calibration loop needs.
-Fix non-calibrated parameters by passing a reference `Monad` instead of `InputFolders`; ModelManager also accepts a `StudySpec`.
+!!! tierwhy
+    [`CalibrationProblem`](@ref) bundles everything the calibration loop needs: which model to run,
+    which parameters vary and under what priors, what was observed, how one simulation is measured,
+    and how that measurement is scored against the observation. Only the first of those has more
+    than one spelling — inputs alone, a reference monad, or a [`StudySpec`](@ref) — and the choice
+    is about what fixes the parameters you are *not* calibrating.
```
drbergman: If you're going to say the choice is about this, then spell out how each choice affects this

drbergman: looks like that happens below, so just say "as explained below"

docs/src/man/calibration.md
```diff
@@ -101,12 +125,47 @@ problem = CalibrationProblem(
...
+```
 
 ### Parameters
 
-Any `DistributedVariation`, `CoVariation{DistributedVariation}`, or `LatentVariation` can be a calibration parameter.
-Priors come from `Distributions.jl`.
+!!! tierwhy
+    Any `DistributedVariation`, `CoVariation{DistributedVariation}`, or `LatentVariation` can be a
+    calibration parameter, and priors come from `Distributions.jl`. A `CoVariation` draws all its
+    members from one CDF value; a `LatentVariation` sends one scalar through user-supplied maps to
```
drbergman: why limit `LatentVariation` to one scalar? It can take several latent parameters

docs/src/man/calibration.md
```diff
@@ -101,12 +125,47 @@ problem = CalibrationProblem(
...
+    several XML paths, which is how a parameter with no XML path of its own gets calibrated.
+
+    **`LatentVariation` needs `inverse_maps` to use the simulation bank.** For
+    `DistributedVariation` and `CoVariation` parameters the inverse maps are constructed
+    automatically; for `LatentVariation` they must be supplied. Without them the simulation bank
+    (`cdf_grid_k`) is **silently disabled** for the entire calibration: proposals cannot be matched
+    to existing monads, so every proposal triggers a new simulation. Supply one inverse map per
+    latent dimension, satisfying `inverse_maps[i](maps[i](u)) ≈ u`. If you do not set `cdf_grid_k`,
+    omitting them is harmless.
```
drbergman: If you don't set them, you can't resume, which is not harmless.

docs/src/man/calibration.md
```diff
@@ -101,12 +125,47 @@ problem = CalibrationProblem(
...
+    Any `DistributedVariation`, `CoVariation{DistributedVariation}`, or `LatentVariation` can be a
+    calibration parameter, and priors come from `Distributions.jl`. A `CoVariation` draws all its
+    members from one CDF value; a `LatentVariation` sends one scalar through user-supplied maps to
+    several XML paths, which is how a parameter with no XML path of its own gets calibrated.
+
+    **`LatentVariation` needs `inverse_maps` to use the simulation bank.** For
+    `DistributedVariation` and `CoVariation` parameters the inverse maps are constructed
+    automatically; for `LatentVariation` they must be supplied. Without them the simulation bank
+    (`cdf_grid_k`) is **silently disabled** for the entire calibration: proposals cannot be matched
+    to existing monads, so every proposal triggers a new simulation. Supply one inverse map per
```
drbergman: Not the best way to put this. "Proposals are not snapped to existing monads, so proposals generate new Monads almost everywhere." It is getting a little mathematical, but what you have sounds like we don't reuse sims if we land right on them

docs/src/man/calibration.md
```diff
@@ -129,32 +189,22 @@ lv = LatentVariation(
...
-    ```
+!!! tierwhy
+    A summary statistic measures **one simulation**; ModelManager combines a parameter set's
+    replicates for you. A plain function is averaged across replicates with `mean`; pass a
+    [`QoI`](@ref ModelManager.QoI) when you need a different reduction, or when the quantity must
+    be computed *after* the replicates are combined rather than before.
 
-    If you do not need the bank (i.e. `cdf_grid_k` is not set), omitting `inverse_maps` is harmless.
+    **Annotate the argument `::Simulation`.** A `summary_statistic` receives a `Simulation`, not a
+    monad ID, and does no averaging of its own. A function written against the monad-level contract
```
drbergman: talking about the monad ID part here is talking about history. a user doesn't need this

docs/src/man/calibration.md
```diff
@@ -165,23 +215,24 @@ end
...
-    erroring if it is untyped, so declare the argument `::Simulation` — ModelManager warns when it
-    is not declared. The built-in measurements below already have that shape.
-
-The built-in measurements are described in [Built-in summary statistics](@ref builtin_ss).
+### Distance functions
 
-The keys of `observed_data` are the comparison: `mseDistance` resolves each one in the simulated summary, an observed key the summary statistic did not produce is an error rather than a silent zero, and an extra simulated component is ignored — a simulation is always known better than the data.
+!!! tierwhy
+    A distance function is any `(simulated, observed) → Float64`. `simulated` is a
+    [`SummaryValues`](@ref): what each QoI's `reduce` returned, keyed by `(qoi name, key)` and
```
drbergman: Give a little more explanation: We use a `SummaryValues` object to support passing in multiple QoIs through compute and reduce and into distance. We should mention (or demo) that if you have a single-valued QoI, e.g. `sim -> 1.0`, then you can get this from `values(simulated) |> first` (or `only`).

docs/src/man/calibration.md
```diff
@@ -165,23 +215,24 @@ end
...
+    QoI reports that key), by the `"<qoi name>.<key>"` label the sink and sensitivity analysis also
+    use (`sim["population_count.cancer"]`), or by the exact tuple
+    (`sim[("population_count", "cancer")]`). A `Real`-valued QoI sits under its name alone.
+    `observed` is whatever you set `observed_data` to.
 
-### Distance functions
+    The keys of `observed_data` are the comparison: [`mseDistance`](@ref) resolves each one in the
+    simulated summary, an observed key the summary statistic did not produce is an error rather
+    than a silent zero, and an extra simulated component is ignored — a simulation is always known
+    better than the data.
```
drbergman: This needs to be distinct from the above comment. Structured this way it feels like it is a continuation of this thought and it makes `mseDistance` feel like **the** distance function. Just put a small section for mseDistance that this block and the tiergloss go into along with a code example applying it to a problem

docs/src/man/calibration.md
```diff
@@ -198,31 +249,84 @@ end
 
 ## Running calibration
 
-### `runABC` — convenience entry point
+!!! tierwhy
+    `run` dispatches on the method, so the verb is the same one that launches a trial or a
+    sensitivity analysis. Algorithm settings belong to the method object — build an
+    [`ABCSMC`](@ref) and keep it if you want the run to be reproducible or the settings reused.
```
drbergman: why does building the ABCSMC object make it reproducible? that's the point of the problem.jld2, right? same with settings reused. the benefit is not about "reuse" but using across multiple problems, not reusing for the same problem. I want it to be clear that if they re-run the same code, they get the same thing (don't say that. this version here says "You must do this if you want that benefit" and it's not true)

docs/src/man/calibration.md
```diff
@@ -242,7 +346,13 @@ All fields have defaults and are specified as keyword arguments:
 
 ### Manual epsilon schedule
 
-Supply a strictly decreasing vector to drive ε by hand instead of letting the algorithm adapt it:
+!!! tierwhy
+    Generation `t` uses `epsilon_schedule[t-1]`, and the schedule takes precedence over
+    `epsilon_quantile`. A hand-written schedule can be far more aggressive than the data supports,
+    so `min_acceptance_rate` is worth setting alongside it as a safety stop.
```
drbergman, line 362: we should add that to the example tehn, the `min_acceptance_rate`

docs/src/man/calibration.md
```diff
@@ -251,37 +361,68 @@ method = ABCSMC(
...
-`min_acceptance_rate` is useful as a safety stop alongside a schedule.
-
 ### Simulation bank and CDF-grid snapping (`cdf_grid_k`)
 
-When `cdf_grid_k = k` is set, ModelManager builds a registry (the *simulation bank*) of all existing monads whose calibrated parameters fall inside the prior support.
-For each proposal, it first checks whether any bank entry falls within the grid cell around that proposal; if so, that monad is reused at its actual coordinates with no new simulation.
-Only when no bank match is found does the proposal get snapped to the nearest dyadic grid point at depth `k`, where a new simulation is run (or an exact match from a previous calibration is reused).
-The grid refines each generation (`k_eff = k + t − 1`), tracking the narrowing posterior.
+!!! tierwhy
+    With `cdf_grid_k = k`, ModelManager builds a registry — the *simulation bank* — of all existing
```
drbergman, line 380: way too late to be explaining this. we need to link to this section above when we reference this or move this explanation up to the first mention

docs/src/man/calibration.md
```diff
@@ -418,24 +546,41 @@ df, weights = posterior(Calibration(42))
...
 
 ### Convergence diagnostics
 
-[`ConvergenceSummary`](@ref) collects per-generation statistics into a table:
+!!! tierwhy
+    Columns: `t`, `max_epsilon_accepted`, `epsilon_threshold`, `acceptance_rate`, `n_accepted`,
+    `ess`, `ess_fraction`, `n_evaluations`. ModelManager 0.9 split the single `epsilon` in two:
+    `max_epsilon_accepted` is the largest distance the generation actually accepted,
+    `epsilon_threshold` the value it ran against, which is `missing` for generation 1 and for
+    generations recorded before the distinction existed.
```
drbergman: This doesn't feel particularly informative. why are telling users the column headers? The rest is then getting in to history which we don't need

docs/src/man/calibration.md
```diff
@@ -495,89 +711,102 @@ data/outputs/calibrations/1/
...
+    [One measurement, one analysis per cell type](@ref gsa_keyed_qoi).
+
+    To analyse a finished monad directly rather than calibrate against it, use
+    [`finalPopulationCount`](@ref) on a `Monad` or `MonadPopulationTimeSeries`: they take a monad
+    and do their own averaging.
+
+    **One reducer everywhere.** No builder on this page defines a `reduce`, so a monad's replicates
+    are averaged by ModelManager's default per-key mean throughout. That default asks the
+    replicates to report the same cell types, which they always do — the keys are the model's
+    declared cell-type roster, and replicates of a monad share a config.
```
drbergman: a lot of this can be cut. I don't really know what a user needs here. You always seem surprised that the cell-type roster is preserved within Monads and Samplings, but that is a given for PCMM. So stop remarking on that. Maybe just the first sentnece (through "per-key mean throughout") is worth saying?

docs/src/man/calibration.md
```diff
@@ -495,89 +711,102 @@ data/outputs/calibrations/1/
...
-
-To analyse a finished monad directly rather than calibrate against it, use
-[`finalPopulationCount`](@ref) on a `Monad` or `MonadPopulationTimeSeries`: they take a monad
-and do their own averaging.
+## Built-in distance functions
 
-[`meanPopulationTimeSeriesQoI`](@ref) does **not**: each of its components is a time series rather
-than the `Real` an index is computed from. See
-[One measurement, one analysis per cell type](@ref gsa_keyed_qoi).
+### [`mseDistance`](@id mse_distance_section)
```
drbergman, line 815: similar to other points: if we have this section way down at the bottom but reference it throughout, we need to at least link to it above

docs/src/man/calibration.md
```diff
@@ -495,89 +711,102 @@ data/outputs/calibrations/1/
...
-    default per-key mean throughout. That default asks the replicates to report the same cell types,
-    which they always do — the keys are the model's declared cell-type roster, and replicates of a
-    monad share a config.
+!!! tierwhy
+    [`mseDistance`](@ref) walks the keys of `observed_data`: each is resolved in the simulated
+    [`SummaryValues`](@ref) (an observed key with no simulated counterpart is an error; extra
+    simulated components are ignored), every squared difference is summed — a time-series value
+    contributing one difference per time point — and the total is divided by the number of
+    differences computed. A series key therefore weighs its whole length against an endpoint key's
+    single term; write your own `distance` to weight them differently, or reduce the series first.
```
drbergman: This last sentence is unclear. Just cut it. The reader will understand what this normalization means.

docs/src/man/calibration.md
```diff
@@ -198,31 +249,84 @@ end
...
-```
+!!! tierdev
+    **What `run` returns.** An [`ABCResult`](@ref) holds `calibration::Calibration` (the database
+    row and its folder), `generations::Vector{GenerationResult}` in order, `parameters` — the
+    [`CalibrationParameter`](@ref) objects that convert CDF coordinates back to target values — and
+    `method::ABCSMC`. The run-level accessors take the result directly, so `.calibration` is
+    optional: `Sampling`, `monadIDs`, `simulationIDs`, `tag!`, `tags`, [`calibrationsTable`](@ref)
+    and [`deleteCalibration`](@ref) all forward to it.
+
+    **The older wrapper names.** Before `run` dispatched on the method, two verbs did this job and
```
drbergman: don't need to give a history lesson

drbergman: cut this whole paragraph

docs/src/man/calibration.md
```diff
@@ -251,37 +361,68 @@ method = ABCSMC(
...
 ```julia
 method = ABCSMC(population_size = 200, max_nr_populations = 10, cdf_grid_k = 3)
 ```
 
+!!! tierdev
+    **Generation 1 is never warm-started.** Seeding it with pre-existing simulations would bias the
+    population away from the prior — prior sweeps and sensitivity designs cluster at particular
+    values — so every gen-1 particle is placed via the Sobol sequence instead.
+    `Monad(...; use_previous=true)` is used internally for every particle regardless, so an exact
+    parameter point already in the database is reused for free.
```
drbergman: MM's `_runFirstGeneration` does do `_lookupAndSnap`. So I don't think this is accurate.

docs/src/man/calibration.md
```diff
@@ -465,17 +615,83 @@ simulation's final cell count on the template project — the four plots look li
...
+# `n_replicates` defaults to the run's; a Calibration works in place of the result
+sampling = createTrial(Calibration(42), draws; n_replicates = 5)
+```
+
+!!! tierdev
+    `createTrial(result, draws)` turns each row's target values into one `DiscreteVariation` per
+    calibrated target, resolved against the run's reference variation exactly as the calibration
+    created its own monads. A plain draw therefore resolves to the monad it came from and adds no
+    simulations unless `n_replicates` exceeds the run's; a smoothed draw creates a new monad. A
+    `Sampling` is a set, so a parameter set drawn more than once appears once — use the `monad_id`
```
drbergman: "sampling is a set is a little confusing. I don't think we even need to mention this since the chance of drawing repeat pars is nearly 0 (except if they are all discrete). I think just drop this note entirely

docs/src/man/calibration.md
```diff
@@ -495,89 +711,102 @@ data/outputs/calibrations/1/
...
-read as they are, and are moved into the folder layout the first time the run is resumed.
-
-ModelManager owns this layout and documents it in full — including what each column means — under
-[What a run leaves on disk](https://drbergman-lab.github.io/ModelManager.jl/stable/man/calibration/#What-a-run-leaves-on-disk).
+!!! tierdev
+    Calibrations written before ModelManager 0.9 stored the same artifacts as flat files
+    (`generation_01.csv`, `generation_01_monads.csv`, and a `generation_cdfs/` subdirectory). Those
+    are read as they are, and are moved into the folder layout the first time the run is resumed.
+    ModelManager owns this layout and documents every column under
+    [What a run leaves on disk](https://drbergman-lab.github.io/ModelManager.jl/stable/man/calibration/#What-a-run-leaves-on-disk).
```
drbergman, line 730: we don't need a history lesson

drbergman (commented):
I would like to see the `Detail` buttons in a more prominent position on the page. Could we try them above the ToC so users are more quick to see them? I do think the bottom is a better long term home, but it will be missed there for a while. They should also be described on the homepage (`docs/src/index.md`). Just saw it's actually there already, gave a comment there.

I agree with the `Getting Started` section not  having the tiers implemented: it's meant for new users who need uniform hand-holding. But I think starting in `Building & Varying Models`, we should start applying the tiers uniformly. Right now, I would like to see the `Code` tier only showing the code and it's not on `xml_path_helpers`

covariations.md did not get any of the tiering. in particular, the "Code" version needs to be significantly more dense.

Why not have the titles of each section be "Brief" "Full" and "Dev" rather than "" "WHY" and "EXTENDING"?

There's only one `tierjournal` even used. I'm sure we can find more things to journal about. I do agree with the one journal there.

docs/journal.jl
```diff
@@ -0,0 +1,80 @@
+# Collects every
+#     !!! tierjournal "YYYY-MM-DD — Title"
+# block under docs/src into docs/src/dev/journal.md, newest first. The blocks
+# stay on their own pages too — this is the same source read in time order.
+
+const JOURNAL_SRC = joinpath(@__DIR__, "src")
+const JOURNAL_OUT = joinpath(JOURNAL_SRC, "dev", "journal.md")
+# ASCII-only on purpose: the dash separator is trimmed below rather than
+# matched here, so the pattern needs no Unicode character class.
+const JOURNAL_HEAD = r"^!!!\s+tierjournal\s+\"(\d{4}-\d{2}-\d{2})\s*(.*?)\"\s*$"
```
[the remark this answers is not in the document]
drbergman: From Claude: Fixed. `journal.jl` now errors on a dated header with no title.

docs/src/assets/tiers.css
```diff
@@ -0,0 +1,153 @@
...
+  display: block !important;
+}
+
+.docs-tier-hidden { display: none !important; }
+
+@media print {
+  .admonition.is-category-tiergloss,
+  .admonition.is-category-tierwhy,
+  .admonition.is-category-tierdev,
+  .admonition.is-category-tierjournal { display: block !important; }
```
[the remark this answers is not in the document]
drbergman: From Claude: Fixed. The print rules repeat the `html[data-docs-tier]` prefix so they match the hide rule's specificity and win on source order.

docs/src/man/calibration.md
```diff
@@ -465,17 +615,83 @@ simulation's final cell count on the template project — the four plots look li
...
+    `Description` — and [`printCalibrationsTable`](@ref) prints it, or sends it to any `sink`.
+    [`deleteCalibration`](@ref) removes a run's database row, its tags, and its output folder.
+
+```julia
+calibrationsTable()                       # every run in the project
+calibrationsTable(; tags = true)          # plus one column per tag key in use
+calibrationsTable([1, 2, 3])              # specific IDs; a Calibration or a run's result also works
+
+printCalibrationsTable()
+printCalibrationsTable(; sink = CSV.write("calibrations.csv"))
```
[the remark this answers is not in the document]
drbergman: From Claude: Changed to an explicit closure, `sink = df -> CSV.write("calibrations.csv", df)`, with `using CSV` above it. For the record the original form also worked: CSV.jl defines `write(file)` to return a one-argument closure. The explicit form is kept because it reads unambiguously.

docs/src/man/post_processing.md
```diff
@@ -1,105 +1,137 @@
...
     [`postProcessingTable`](@ref), each populated only for the runs that produced it.
 
-[`populationFractionQoI`](@ref) is the same builder for each cell type's share of the population, and works here as well as in calibration and sensitivity analysis. See [QoI form](@ref qoi_form_ss).
-
 ## Reading the stored quantities back
 
 ```julia
 postProcessingTable(sampling)                     # just the stored quantities, one row per simulation
+printPostProcessingTable(sampling)                # the same table, printed
+printPostProcessingTable(sampling; sink=CSV.write) # ...or sent somewhere else
```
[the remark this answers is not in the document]
drbergman: From Claude: Fixed. This one was genuinely broken (a bare `CSV.write` gets only the table). Now `using CSV` plus `sink = df -> CSV.write("qois.csv", df)`.

docs/src/man/trials.md
```diff
@@ -0,0 +1,109 @@
...
+
+```julia
+simulationsTable(sampling)
+printSimulationsTable(sampling)
+
+monadsTable(sampling)
+printMonadsTable([monad, sampling])
+
+using CSV
+printSimulationsTable(sampling; sink = CSV.write("runs.csv"))
```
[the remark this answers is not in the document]
drbergman: From Claude: Changed to `sink = df -> CSV.write("runs.csv", df)`. The original was valid (`CSV.write(file)` returns a closure) but the explicit form is clearer.

docs/src/dev/architecture.md
drbergman: From Claude: Added `run(inputs, variations; n_replicates=3)` as a second line of that block, with a sentence saying it is the form most scripts use and that the walkthrough applies to both.

docs/src/dev/simulator_interface.md
drbergman: From Claude: Fixed, and swept every page in the PR for British spellings (fulfil, centre, labelled, honour, analyse, materialise, initialise, neighbour). CONTRIBUTING.md now says American spelling throughout.

docs/src/dev/simulator_interface.md
drbergman: From Claude: Rewritten as "the foreign-key column that `simulations`, `monads` and `samplings` each carry, pointing at a row of the simulator version table".

docs/src/dev/utilities.md
drbergman: From Claude: You are right: `tag::child_tag:content`, two colons (`src/configuration.jl:37`). Fixed the text and added a fourth line to the example showing the `::` form.

docs/src/man/analyzing_output.md
drbergman: From Claude: Added. The gloss now says each panel's title is the parameter values that distinguish that monad from the others in the sampling, and a Full block explains it is that monad's row of `simulationsTable(sampling)` minus the ID columns, which is also the panel order (verified in `src/analysis/population.jl`).

docs/src/man/analyzing_output.md
drbergman: From Claude: Done. The code block now shows what each loader fills in (`snapshot.cells` with cell-type names, `snapshot.substrates` with one named column per substrate, `snapshot.mesh`, `snapshot.neighbors`), and the gloss says which function reads which names from the XML (`loadCells!` and `loadSubstrates!`; the other two take none).

docs/src/man/analyzing_output.md
drbergman: From Claude: Removed the subsection. `include_dead` and `dr` are `pcf`'s own keywords and are now two clauses at the end of its gloss, the same way every other function on the page treats keywords.

docs/src/index.md
drbergman: From Claude: Done: a `## How much detail do you want?` heading and a five-row table (level, what you see), plus a line that the Getting Started pages read the same at every level.

docs/src/man/varying_parameters.md
drbergman: From Claude: Cut. `shortVariationName` now appears in one sentence saying what it returns, with no advice about other simulators. That rule (PCMM docs address PCMM users and developers, not people building on ModelManager) is in CONTRIBUTING.md.

docs/src/man/varying_parameters.md
drbergman: From Claude: Done. Code tier is now the lead sentence, six headings and seven code blocks; the three XML-path snippets are one block and the distributed shorthands are one block.

docs/src/man/varying_parameters.md
drbergman: From Claude: Done. Code: the five `createTrial` calls. Brief: one line per method. Full: a paragraph per method with its signature, each linking its docstring. I dropped the claim that Sobol' wants `n` near a power of two, since neither `SobolVariation`'s nor `addVariations`' docstring says so.

docs/src/man/varying_parameters.md
drbergman: From Claude: Moved. `domainVariations` has its own `## Domain bounds in bulk` section at the end of the page (heading and code visible, explanation in the Dev block).

docs/src/man/latent_variations.md
drbergman: From Claude: Moved to a Full block (`tierwhy`) with the qualified `@ref` kept.

docs/src/man/latent_variations.md
drbergman: From Claude: Done. The three prose-only headings are gone; Code tier is the lead sentence, one naming call, and the two doctests under the two type headings.

docs/src/man/intracellular_inputs.md
drbergman: From Claude: Done: Code tier is the lead sentence and one eight-line block (assemble, read the assigned ID, build the XML path, pass the folder to `InputFolders`). While there, the example was wrong in the way Copilot's suppressed comment described: with a `Dict{String,PhysiCellComponent}` the caller's `component.id` stays `-1`. It now builds `Dict(cell_type => [component])` and reads `cell_type_to_components[cell_type][1].id`, matching `IntracellularTests.jl`.

docs/src/man/running_simulations.md
drbergman: From Claude: Added to the Brief block beside `setJobOptions`: PCMM sets each job's `cpus-per-task` from that simulation's `omp_num_threads`, a value you pass replaces it and is never overridden, and an under-allocated job time-slices its threads with nothing in the logs to say so. A journal entry (2026-09-05) records the decision.

docs/src/man/running_simulations.md
drbergman: From Claude: Agreed. The four-field walk-through is now a Full block; the Brief keeps one line.

docs/src/man/trials.md
drbergman: From Claude: Done everywhere, not just here: every adjacent why/gloss pair in the manual was swapped to gloss then why, and CONTRIBUTING.md records the order as gloss, why, code, dev, journal. Note this is the opposite of the `julia-tiered-docs` skill's order, so the skill may want the same change.

docs/src/man/trials.md
drbergman: From Claude: Added at Full. One nuance from the `monadIDs` docstring: a `Simulation(inputs, variation_id)` constructed directly resolves to the monad sharing its parameters but is only listed among that monad's replicates once `run` records it. The text says the monad is where a simulation is recorded and `run` is what puts it there, so there is no simulation without a monad above it.

docs/src/man/trials.md
drbergman: From Claude: Changed to "to see every parameter the database holds for those rows".

docs/src/man/trials.md
drbergman: From Claude: It does not sort by them: tag columns (and post-processing columns) are appended as-is and excluded from sorting, per the `simulationsTableFromQuery` docstring. The text now says so and points at `sort_by` for choosing the sort.

docs/src/man/sensitivity_analysis.md
drbergman: From Claude: Cut. Now "These are usually `DistributedVariation`s".

docs/src/man/sensitivity_analysis.md
drbergman: From Claude: Reduced to "a sensitivity function must accept a `Simulation`", plus the value-shape rule and a link to the `QoI` docstring. `CalibrationProblem`'s docstring no longer mentions a warning for an unannotated argument, so that advice is gone too.

docs/src/man/sensitivity_analysis.md
drbergman: From Claude: Removed the section. The one thing it carried that matters (a `Dict`-valued QoI gives one analysis per key) is a line in the "Running the analysis" gloss with the `functions=[populationCountQoI()]` call in that code block; `@id gsa_keyed_qoi` moved there so the calibration page's link still resolves.

docs/src/man/sensitivity_analysis.md
drbergman: From Claude: Gone with the section.

docs/src/man/sensitivity_analysis.md
drbergman: From Claude: Agreed; QoI naming is not explained on this page anymore. It links the `QoI` docstring instead.

docs/src/man/sensitivity_analysis.md
drbergman: From Claude: Rewritten along those lines: results accumulate on the sampling and a present label is skipped; the stored results are what makes adding a quantity cheap, with a link to the post-processing page for where computed quantities are kept; `recompute=true` evaluates again instead of reusing them.

docs/src/man/calibration.md
drbergman: From Claude: `scale[min(t, end)]`: the last entry holds for every later generation. The code comment and the kernels block now say so (from the `GaussianKernel` docstring).

docs/src/man/calibration.md
drbergman: From Claude: Done. `run(calibration)` is the resume call in every example, including `run(calibration; problem = ...)` and the keyword patch. `run(::Calibration)` forwards its keywords to `resumeCalibration`, so nothing needs the older name; `resumeCalibration` and `resumeABC` appear once at the end of the block as equivalents.

docs/src/man/calibration.md
drbergman: From Claude: Changed to `sim -> finalPopulationCount(sim)["cancer"] / 1000`.

docs/src/man/calibration.md
drbergman: From Claude: `samplePosterior` has its own section, "Sampling the posterior", and "Posterior samples" now links to it.

drbergman: From Claude: Linked (`@ref posterior_sampling_calibration`).

docs/src/man/calibration.md
drbergman: From Claude: Agreed, but that is ModelManager's corner-plot recipe, not something this PR can change. Noted in progress.md as a ModelManager follow-up.

docs/src/man/calibration.md
drbergman: From Claude: Done: four code blocks, each directly above the figure it produces.

docs/src/man/calibration.md
drbergman: From Claude: Added a Brief block: the bank lets a calibration reuse monads that already exist in the project instead of simulating every proposal afresh.

docs/src/man/calibration.md
drbergman: From Claude: It really is a tuple, `(df, weights)`; that is ModelManager's `posterior`. The page states it plainly. Whether ModelManager should return one frame with a weight column is a ModelManager decision; noted in progress.md as a follow-up.

docs/src/man/calibration.md
drbergman: From Claude: Renamed to "Using the builders in sensitivity analysis" (the `@id qoi_form_ss` is kept because the post-processing page links it, and that link's text was updated to match).

docs/src/man/calibration.md
drbergman: From Claude: Moved into a Dev block, and the reasoning is a journal entry (2026-04-22, a Julia-native ABC-SMC rather than a bridge to a Python one).

docs/src/man/calibration.md
drbergman: From Claude: Now ends "as explained below".

drbergman: From Claude: Done.

docs/src/man/calibration.md
drbergman: From Claude: Fixed: "sends its latent parameters, one or several, through user-supplied maps".

docs/src/man/calibration.md
drbergman: From Claude: Rewritten: leaving `inverse_maps` out means proposals are not snapped to existing monads, so nearly every proposal creates a new monad, and the run cannot be resumed. No "harmless".

docs/src/man/calibration.md
drbergman: From Claude: Rewritten as above; "silently disabled" is gone.

docs/src/man/calibration.md
drbergman: From Claude: Cut the whole paragraph.

docs/src/man/calibration.md
drbergman: From Claude: Added: `SummaryValues` is the object that carries each QoI's reduced value from `compute` through `reduce` into `distance`, so one problem can measure several quantities at once; for a single-valued QoI the value is `only(values(simulated))`, shown in the code block.

docs/src/man/calibration.md
drbergman: From Claude: The key-resolution text moved into the existing `mseDistance` section, which now has a Brief, a Full, and a `CalibrationProblem` example applying it. "Distance functions" links to that section at the first mention.

docs/src/man/calibration.md
drbergman: From Claude: Removed the reproducibility claim. It now says settings live on the method object, and keeping the object in a variable is worth doing when you want the same settings for several problems.

docs/src/man/calibration.md
drbergman: From Claude: Added `min_acceptance_rate = 0.01` to the example.

docs/src/man/calibration.md
drbergman: From Claude: The first mention of the bank (in the Parameters block) and the `cdf_grid_k` row of the settings table now link to the bank section.

docs/src/man/calibration.md
drbergman: From Claude: Cut the column listing and the history. It now says what `ConvergenceSummary` is for (telling a run that converged from one that merely stopped) and points at the docstring for the columns.

docs/src/man/calibration.md
drbergman: From Claude: Reduced to that first sentence.

docs/src/man/calibration.md
drbergman: From Claude: Linked at the first mention.

docs/src/man/calibration.md
drbergman: From Claude: Cut.

docs/src/man/calibration.md
drbergman: From Claude: Cut. Both names remain in one Dev-block sentence stating they are equivalent to `run(method, problem)`, which the coverage test needs.

drbergman: From Claude: Done.

docs/src/man/calibration.md
drbergman: From Claude: Removed the paragraph. The bank block now states only eligibility, the index, and that discrete-prior coordinates are never snapped, all from the `ABCSMC` docstring.

docs/src/man/calibration.md
drbergman: From Claude: Dropped.

docs/src/man/calibration.md
drbergman: From Claude: Dropped; the block keeps only the ModelManager "What a run leaves on disk" link.

drbergman:
From Claude: pushed 2dfcee9 addressing the review. Per point in the review body:

- **Detail selector position:** it now mounts above the sidebar's table of contents (before `ul.docs-menu`), and the home page has a `## How much detail do you want?` heading with a table of the five levels.
- **Getting Started untiered, tiers uniform from Building & Varying Models on:** applied. The six pages that had kept bare prose (xml_path_helpers, varying_parameters, covariations, latent_variations, intracellular_inputs, querying_parameters) and the reference pages (best practices, data directory, project configuration, known limitations, examples, movies, Studio) are tiered; at Code tier each is its lead sentence, headings and code. Known limitations has no code, so its Code tier is the four headings, which I left as is rather than invent code.
- **Marker titles:** the small markers now read "brief" / "full" / "dev" (the tier that reveals the block) instead of "why" / "extending".
- **Gloss before why:** every adjacent pair was swapped, and CONTRIBUTING.md records the order gloss, why, code, dev, journal. This is the reverse of the `julia-tiered-docs` skill's order, so the skill may want the same change.
- **More journal entries:** 25 now, mined from progress.md and dated from it (calibration 4, post-processing 3, analyzing output 3, dev pages 6, the rest one or two each).

Two things the review raised that belong to ModelManager, noted in progress.md: `posterior` returning `(df, weights)` rather than one frame, and the corner plot's empty upper-triangle axes.

Every inline comment has a reply saying what changed. Unreviewed pages got the same treatment; the extrapolations I would double-check are `managing_a_project.md` (the renamed-names block is now three bare aliases) and `examples.md` (its `runABC(problem)` became `run(ABCSMC(...), problem)`).

docs/src/man/running_simulations.md
```diff
@@ -52,36 +52,42 @@ pathToOutputFolder(Simulation(5))
...
-!!! tiergloss
-    [`setNumberOfParallelSims`](@ref) caps how many simulations run concurrently on this machine.
-    It starts from the `PCMM_NUM_PARALLEL_SIMS` environment variable, or 1 if that is unset.
-
 ```julia
 setNumberOfParallelSims(8)
 run(sampling)
 ```
 
 ## Run on an HPC
```
drbergman: We need to state here how PCMM detects if it's on a SLURM HPC, where it displays that at init and how `useHPC` can be used to set this.

drbergman: You have it in Full, but this is essential and belongs in Brief. It's about how the HPC is even used in the first place and how the user can detect that PCMM is going to submit jobs

docs/src/man/running_simulations.md
```diff
@@ -90,30 +96,51 @@ setJobOptions(Dict("time" => "02:00:00",
...
+    page has the rest of the hooks.
+
+!!! tierjournal "2026-09-05 — `cpus-per-task` follows `omp_num_threads`"
+    **Decided:** PCMM answers ModelManager's `simulationThreads` hook with
+    `parallel/omp_num_threads` read from the simulation's own variation record, so a varied thread
+    count is honored and every job is allocated the CPUs PhysiCell will actually use. A config
+    whose element cannot be read falls back to 1 with one warning rather than failing the
+    submission, and a `cpus-per-task` you set yourself still wins.
+    **Rejected:** installing the default from PCMM as a `Function` job option, which is the same
+    mechanism with a wrapper and does not let other backends fill the default the same way.
```
drbergman: cut. too similar to above dev

docs/src/man/running_simulations.md
```diff
@@ -90,30 +96,51 @@ setJobOptions(Dict("time" => "02:00:00",
...
+    `submit_retry_period` (how long a transiently refused submission is retried), `poll_interval`
+    (how often a worker checks for its own sentinel), `reap_interval` (how long one `squeue` answer
+    is shared among waiting workers), and `grace_period` (how long a job may be missing from the
+    queue with no sentinel before it is declared failed).
+
     `grace_period` is the one to raise first. It must exceed your filesystem's worst-case
     directory-attribute staleness: if a compute node writes the sentinel but this node cannot see
     it within the grace period, a successful simulation is recorded as failed. The sentinel
     directory itself is deliberately not one of these options — it is `data/outputs/.hpc_done`
     unless `MODELMANAGER_HPC_DONE_DIR` says otherwise, and it is fixed for the session.
```
drbergman: We need to state the environment variable more clearly than this

docs/src/man/trials.md
```diff
@@ -12,17 +12,22 @@ how to get one back, and how to read what it contains.
...
 
+!!! tierwhy
+    You do not choose the class; you describe the runs and the class follows. That is what lets a
+    script grow a `DiscreteVariation` from one value to three without any other edit — the same
+    call that returned a `Simulation` now returns a `Sampling`, and [`run`](@ref) takes either.
+
+    A `Simulation` is never on its own, either. The `Monad` for its parameter set is where a
+    simulation is recorded, and [`run`](@ref) is what puts it there, so a single run and a
+    five-replicate monad differ only in how many simulations that monad holds — there is no
+    simulation without a monad above it. [`monadIDs`](@ref) on a simulation names that monad.
```
drbergman: Even if the `createTrial` call created a `Simulation`, that `Simulation` is added to a `Monad` once it is run. Future `Simulation`s with the same inputs and parameters are automatically added to that `Monad`.

or even simpler

Every `Simulation`, once run, is guaranteed to belong to a `Monad`.

docs/src/man/calibration.md
```diff
@@ -10,11 +10,21 @@ the posterior.
...
 
-    **What comes from where.** The implementation is native Julia — no Python or conda environment
-    is required. The algorithm lives in ModelManager; PhysiCellModelManager.jl contributes the
-    PhysiCell-specific summary statistics ([`populationCountQoI`](@ref),
-    [`populationFractionQoI`](@ref), [`meanPopulationTimeSeriesQoI`](@ref)). One
-    `using PhysiCellModelManager` brings both.
+!!! tierdev
+    The implementation is native Julia and needs no Python or conda environment. The algorithm is
+    ModelManager's; PhysiCellModelManager.jl contributes the PhysiCell-specific summary statistics,
+    and one `using PhysiCellModelManager` brings both.
```
drbergman: This is history and belongs just in journal. At most, a dev would want to konw this is from MM, but that's true of much more in these docs that we don't state explicitly.

docs/src/man/calibration.md
```diff
@@ -217,30 +222,34 @@ problem = CalibrationProblem(inputs, params, observed, my_stat, mseDistance)
 
 ### Distance functions
 
+!!! tiergloss
+    A distance is any `(simulated, observed) → Float64`.
+    [`mseDistance`](@ref mse_distance_section) is the built-in one; a custom function may use any
+    types.
```
drbergman: I think the "a custom function may use any types." part is referring to the `observed` argument?

docs/src/man/calibration.md
```diff
@@ -550,26 +569,23 @@ df, weights = posterior(Calibration(42); generation = 3)
     A [`GenerationResult`](@ref) is what a generation actually stored, and `posterior` is the view
     of it in target space. Its `particles` frame holds **latent CDF coordinates**, the algorithm's
```
drbergman: what is "Its" here? The `GenerationResult` hold `particles?

docs/src/man/calibration.md
```diff
@@ -653,35 +668,34 @@ sampling = createTrial(Calibration(42), draws; n_replicates = 5)
...
-
 ```julia
 calibrationsTable()                       # every run in the project
 calibrationsTable(; tags = true)          # plus one column per tag key in use
 calibrationsTable([1, 2, 3])              # specific IDs; a Calibration or a run's result also works
 
 printCalibrationsTable()
-printCalibrationsTable(; sink = CSV.write("calibrations.csv"))
+using CSV
+printCalibrationsTable(; sink = df -> CSV.write("calibrations.csv", df))
```
drbergman: why not use the simpler version we had before? `sink = CSV.write("calibrations.csv")`?

docs/src/man/calibration.md
```diff
@@ -756,53 +778,54 @@ populationFractionQoI(; index=:final, cell_types=nothing, include_dead=false)
...
 
+!!! tierjournal "2026-09-14 — One builder per quantity, one reducer for all of them"
+    **Decided:** each quantity gets exactly one builder — `populationCountQoI(; index)` and
+    `populationFractionQoI(; index)` read the snapshot at `index`, `:final` by default — and none of
+    them defines a `reduce`, so all are averaged by ModelManager's default per-key mean. Sink
+    columns and labels are `population_count.<cell_type>` and `population_fraction.<cell_type>`.
+
+    **Rejected:** a separate endpoint builder beside each. Once the two reduced identically it was
+    the same measurement under a second name and a second family of sink columns.
+
```
drbergman: why is this in the calibration page? This should be on the sensitivity page or on the QoI page.

